PR #31, opened 2024-09-09
TMRG cmake function udpate

benoitdenkinger:
TMRG cmake function updated with:

1. by default the --no-header option is used unless the HEADER argument is passed
2. To keep the header timestamp (to have a simple way to check when were generated the triplicate files) the date and time is added to the stamp file content.

This way, tracking triplicated files and checking for changes is simpler (otherwise files always change because of the different time). It also enable sanity check using CI to see if files copied using the copy_rtl cmake function have been manually modified.


benoitdenkinger:
@mksoc I know you wanted to keep the tmrg header to check for the generation time. I propose an alternative so when we track tmrg generate files they do not change every time and checking the history with git is simpler.

PR #45, opened 2024-09-12
Add support for vhier, generating file list, module list, xml file

Risto97:
I liked the idea of `copy_rtl_files()`, however I would separate `vhier` as a separate target and copy_rtl_files should process `xml` file generated from `vhier`.

This function also uses Generator expressions for handling function arguments, which is the proposed way of handling function arguments


mksoc:
At the moment, I'm actually using a modified version of `vhier` to print a more useful output for us (https://github.com/HEP-SoC/verilog-perl), but if we parse the XML I believe we already have everything we need, even with the "stock" `vhier`. 

We can look into modifying the Python script to parse the XML and change `copy_rtl_files()` accordingly.

PR #48, opened 2024-09-13
sed_wor dedicated function

nookfoo:
extracted `sed_wor` from `verilate.cmake` into dedicated `sed_wor.cmake`.  See PR #40 for more info.


Risto97:
Looks good.
However I feel like you didn't need this additional target, and you could have used the one in `tmrg()`

From what I wrote in the comment before:

> In any case I think the fact that the tmrg function is applying sed might not be problem for you.
> If the structure of triglav is mostly the same as before, your verification and implementation directories are different.
> Meaning when you do implementation you configure CMake there, and there is no verilator target.
> 
> In that case you can have a variable that will affect if tmrg function applies sed or not, ie. when you do implementation don't apply, when simulation apply...
> Something like
> 

```
 tmrg(${IP} $<$<BOOL:${SIMULATION}>:SED_WOR>)
```

Also for @benoitdenkinger 
I like the idea of adding `date` to the timestamp file.
I am curious for what are you using `date` in the timestamp file?

nookfoo:
> Looks good. However I feel like you didn't need this additional target, and you could have used the one in `tmrg()`

Yeah, I think there were some files which are triplicated code but are not being passed through `tmrg` (some adapters), so I wasnt sure about tieing it to `tmrg` target and made a new one.

> 
> From what I wrote in the comment before:
> 
> > In any case I think the fact that the tmrg function is applying sed might not be problem for you.
> > If the structure of triglav is mostly the same as before, your verification and implementation directories are different.
> > Meaning when you do implementation you configure CMake there, and there is no verilator target.
> > In that case you can have a variable that will affect if tmrg function applies sed or not, ie. when you do implementation don't apply, when simulation apply...
> > Something like
> 
> ```
>  tmrg(${IP} $<$<BOOL:${SIMULATION}>:SED_WOR>)
> ```
> 
> Also for @benoitdenkinger I like the idea of adding `date` to the timestamp file. I am curious for what are you using `date` in the timestamp file?

I think it makes more sense to keep to sed_wor with verilator, as it is solely due to verilator issue. We might also have simulation with xrun at RTL level, which would then need another switch between verilator simulation or xcelium simulation.

benoitdenkinger:
> Looks good. However I feel like you didn't need this additional target, and you could have used the one in `tmrg()`
> 
> From what I wrote in the comment before:
> 
> > In any case I think the fact that the tmrg function is applying sed might not be problem for you.
> > If the structure of triglav is mostly the same as before, your verification and implementation directories are different.
> > Meaning when you do implementation you configure CMake there, and there is no verilator target.
> > In that case you can have a variable that will affect if tmrg function applies sed or not, ie. when you do implementation don't apply, when simulation apply...
> > Something like
> 
> ```
>  tmrg(${IP} $<$<BOOL:${SIMULATION}>:SED_WOR>)
> ```
> 
> Also for @benoitdenkinger I like the idea of adding `date` to the timestamp file. I am curious for what are you using `date` in the timestamp file?

Nothing fancy :) I added it to the tmrg target to have the generated files timestamp for debugging purposes (mainly checking when were generated the files). Before the tmrg output files contained the tmrg header (with a timestamp) by default but now the default is no header so when we track tmrg files (yes we do it in implementation repos ;)) their content change only if the generated RTL changes (not because of the new timestamp).

benoitdenkinger (approved):

benoitdenkinger (approved):

PR #56, opened 2024-09-14
allow add_ip short form, resolves #42

### LINKED ISSUE #42: Consider shortened argument for add_ip
Right now `add_ip()` function has a following format:

```
add_ip(uart
    VENDOR company
    LIBRARY ip
    VERSION 0.0.1
    )
```

It is a bit verbose.
It would be possible to have a shortened version:

```
add_ip(company::ip::uart::0.0.1)
```

P.S.
There is already a possibility to omit VENDOR, LIBRARY, and VERSION.
However this would issue a warning.
```
add_ip(uart)
```

What are your thoughts @mksoc @benoitdenkinger @nookfoo ?

Risto97:
This pull request implements the short signature for `add_ip()` function.

Now there are 3 versions:

## 1. Full form
```
add_ip(ip
    VENDOR vendor
    LIBRARY lib
    VERSION 0.0.1
)
```
This translates into `vendor::lib::ip::0.0.1` target

In this version, it is possible to omit any of the VENDOR, LIBRARY, VERSION arguments.
A special case of full form is when all of them are omitted.
```
add_ip(ip)
```
This translates into `ip` target

## 2. Short form

In the short form the VLNV format can be passed to `add_ip()`.
Like this:
```add_ip(vendor::lib::ip::0.0.1)```

In this form it is important that the full form is passed, as if one of the tokens are missing, how can we know what is VENDOR, LIBRARY, IP, VERSION

### DIFF
--- a/cmake/hwip.cmake
+++ b/cmake/hwip.cmake
@@ -9,10 +9,32 @@ include("${CMAKE_CURRENT_LIST_DIR}/utils/safe_get_target_property.cmake")
 # This function is a wrapper around the cmake built-in
 # `add_library() <https://cmake.org/cmake/help/latest/command/add_library.html>`_ function.
 # It generates the library name using the vendor, library, name, and version (VLNV) information passed
-# in arguments (see get_ipname()). It creates two alias libraries to the default <vendor>__<library>__<name>__<version>:
-#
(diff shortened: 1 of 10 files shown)


Risto97:
resolve #42
close #42

Issue #63, opened 2024-09-15
Unit tests for help(), help_targets(), help_ips()


Risto97:
Add unit tests for the following functions

Risto97:
It was not resolved, this is a mistake

Issue #61, opened 2024-09-15
Print a help message


Risto97:
Currently, the help provided with CMake is quite basic, as it only lists the targets without any description.
It is still useful to keep the native `make help` for purposes of bash autocompletion.

However it is possible to have more elaborate help messages for targets and IPs, that would show target name and description in a coloured formatted table

Issue #64, opened 2024-09-15
Add DESCRIPTION property to all the tool targets


Risto97:
To be able to print help information on IP blocks, allow DESCRIPTION argument to all the backend tool functions.

DESCRIPTION property should be assigned to UTILITY (custom_target) target containing the description.

Issue #62, opened 2024-09-15
Add DESCRIPTION argument to add_ip() function


Risto97:
To be able to print help information on IP blocks, allow DESCRIPTION argument to `add_ip()` function.

DESCRIPTION property should be assigned to `INTERFACE_LIBRARY` target containing the description.

PR #65, opened 2024-09-15
add help(), help_targets(), help_ips() functions for help messages

Risto97:
This pull request adds functions:

* `help_targets()` - Prints all targets and their description in the build
* `help_ips()`         - Prints all IPs and their descriptions.
* `help()`                - Prints both targets and IPs help messages

2024-09-15T10:52:59,213558953+02:00

To make this useful, all the tool functions should set `DESCRIPTION` property on the targets it creates for 2nd column to be useful. (mentioned #64 )
`add_ip()` should also have `DESCRIPTION` optional argument (mentioned #62 ) in order to have useful information.

These functions can be added only once per build flow.
It is recommended to add it at the end of the CMakeLists.txt file.

In order to ensure it is the only `help()` call, all levels of hierarchy should have a "top level guard" around the call:

```
if(PROJECT_IS_TOP_LEVEL)
    help()
endif()
```


Risto97:
resolve #63

Risto97:
Resolve #61

Issue #70, opened 2024-09-15
Remove verilate_xml and tmrv


Risto97:
These functions are not useful.
XML format is deprecated in verilator
TMRV never worked and its too specific.

To reduce maintenance workload remove these functions.

Issue #68, opened 2024-09-15
Move lds_gen to separate repo


Risto97:
Linker script generator lds_gen should be its own repository and moved from SoCMake, only the `.cmake` function should stay.

Issue #67, opened 2024-09-15
Remove socgen library from SoCMake


Risto97:
Socgen library should not be in SoCMake, it should be moved on its own repository, or within another repository

PR #72, opened 2024-09-15
Remove unused or overly specific functions

Risto97:
In order to have a repository with only tested and easy-to-maintain tool support, I prefer to have release 1.0.0 come with a minimal set of tool support with higher quality.
Later any of the removed tools can be added following the same standards and structure (argument convention, ...), rather than refactoring many files prior.


Risto97:
resolve #70

Risto97:
resolve #67

Risto97:
resolve #68

Issue #34, opened 2024-09-10
Consecutive Verilate targets bug


Risto97:
Currently, there is a bug that makes it impossible to have consecutive `verilate()` functions.
For some reason on the first call `verilate` function gets overwritten by `verilate` function from Verilator.

Easiest solution is to rename `verilate` to `verilator` or similar.

PR #74, opened 2024-09-15
Iverilog verilator refactor

Risto97:


Risto97:
Add verilator, vhier and iverilog for  #64.
Resolves  #34, need more testing.

PR #50, opened 2024-09-13
Add possibility to list languages in get_ip_sources/include_directori…

Risto97:
Implement capability of getting properties for multiple languages in one call for  `get_ip_sources()`, `get_ip_include_directories()`, `get_ip_compile_definitions()`.

It can be used as following:

```
get_ip_sources(SOURCES ip1 VERILOG SYSTEMVERILOG VHDL)
get_ip_include_directories(INCDIRS ip1 VERILOG SYSTEMVERILOG VHDL)
get_ip_compile_definitions(COMPDEFS ip1 VERILOG SYSTEMVERILOG VHDL)
```

The variable `SOURCES` will contain list of verilog, systemverilog and vhld sources in the order of parameters.
The same is valid for 2 other functions.

This in turn deprecates:
* `get_ip_rtl_sources()`
* `get_ip_sim_only_sources()`
* `get_ip_fpga_only_sources()`

As they didn't provide enough control, i.e. when only Verilog and Systemverilog but not VHDL is needed.
For sim and fpga source it can be done as following:

```
get_ip_sources(SOURCES ip1 VERILOG_SIM SYSTEMVERILOG_SIM VHDL_SIM)
get_ip_sources(SOURCES ip1 VERILOG_FPGA SYSTEMVERILOG_FPGA VHDL_FPGA)
```

Tests are added for new changes

P.S. `NO_DEPS` still works as before and will not be considered as a language:

```
get_ip_sources(SOURCES ip1 VERILOG SYSTEMVERILOG VHDL NO_DEPS)
```

PR #94, opened 2024-10-17
SYSTEMVERILOG_TB source file type

benoitdenkinger:
### SYSTEMVERILOG_TB FILES

Added a new type of source file called SYSTEMVERILOG_TB. It is meant to list files that are not supposed to be synthesized and used by the testbench (e.g., DPI modules, models).

A new function is available to retrieve these files:
get_ip_tb_only_rtl_sources()

This function call has been added to all the simulators.

### ADD_TESTS()

The add test function has been made more generic to be able to pass any executable (being a target or an external one). This enable to use this function and the created check target with any simulation tool (e.g., verilator, xcelium, vcs).

### DIFF
--- a/cmake/sim/verisc/verisc_install.cmake
+++ b/cmake/sim/verisc/verisc_install.cmake
@@ -63,12 +63,6 @@ macro(verisc_build)
         message("veriSC_VERSION ${veriSC_VERSION} not matching requested version ${VERSION}")
     endif()
 
-    if(EXISTS ${VERISC_HOME}/veriSCConfig.cmake)
-        message("veriSCConfig.cmake found")
-    else()
-        message("veriSCConfig.cmake NOT found")
-    endif()
-
     set(VERISC_HOME "${veriSC_DIR}/../../../")
 
     if((NOT veriSC_FOUND) OR (NOT veriSC_VERSION VERSION_EQUAL ${VERSION}) OR FORCE_UPDATE)


Risto97:
What is the difference between _TB sources and _SIM sources?
https://github.com/HEP-SoC/SoCMake/blob/4a413900d16f0af6c470529fce1306e20ed47f48/cmake/hwip.cmake#L264-L276

benoitdenkinger:
When you have a SystemVerilog testbench you pass it as the top module to you simulation tool (e.g., xcelium). In this case, when you get the sources of your design the files you always want are the RTL sources and the TB files, while SIM files are only used when running technology independent simulations.
This is used when you have for example some test modules in your testbench (e.g., to connect interfaces like UART). These modules you always want to fetch them to build your testbench. For SIM files you can have, for example, a generic memory macro that you use for fast prototyping but once you go for an implementation you use a technology memory macro so you don't fetch the SIM files anymore while you still fetch the TB ones.

mksoc:
I'm still not conviced... I feel there's a 90% overlap between _SIM and _TB and maybe we should find a better way to handle this particular case (maybe with additional sources attributes like we discussed some time ago). For the moment it's fine, but let's keep it in mind, because this is not going to be clear to potential users.

Risto97:
I also feel like there are different ways of handling this.
Even though I agree this way is more straight forward in some cases, I don't like that it introduces more complexity into SoCMake, while I feel like there are already few mechanisms that can address this.
Especially the tool backend functions become more complicated.

### Option 1, CMake options to configure the project

I think your suggestion would be better in case you need to have both of the file sets in the current configuration of the project.
Example if you want to run both post synthesis simulation and behavioral simulation at the same time (having 2 targets).
I don't feel like you would ever to this?

If not, you can always drive your configuration with [new option functions](https://github.com/HEP-SoC/SoCMake/pull/76/files).
Example:

```cmake
option_enum("SIM_TYPE" "Type of simulation to run" "SYNTH;GENERIC" "GENERIC")

if(SIM_TYPE STREQUAL SYNTH)
    # ip_sources(${IP} SYSTEMVERILOG <SYNTH_FILES>)
elseif(SIM_TYPE STREQUAL GENERIC)
    # ip_sources(${IP} SYSTEMVERILOG <SIM_FILES>)
endif()

help()
```

You can even get help on the options with `make help_options` and get this print:
2024-10-27T09:38:24,418592177+01:00

To set the options you can pass them during configuration phase with `cmake -DSIM_TYPE=SYNTH <source_dir>`.
Or with `cmake-gui <source-dir>` to see and configure them in GUI interface.

Because you might have a lot of options to pass and different configurations, you might want to use [`cmake-presets`](https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html).
I haven't explored them yet, but they can help when you have different build flows of the design (e.g. synthesis, simulation, post-implementation simulation, etc...).

### Option 2. use `ip_compile_definitions()` and `ifdef`

Another option that can maybe help excluding files and parts of the files from different flows is `compile_definitions`

In Verilog files you can have:

```Verilog
`ifdef SYNTH
    mem_macro mem_inst(...);
`elseif GENERIC
    generic_mem mem_inst(...);
`endif
```

To define the macro for the IP block use the `ip_compile_definitions()` function:
```cmake
option_enum("SIM_TYPE" "Type of simulation to run" "SYNTH;GENERIC" "GENERIC")

add_ip(test)

ip_sources(test SYSTEMVERILOG
    ....
    )

if(SIM_TYPE STREQUAL SYNTH)
     ip_compile_definitions(${IP} SYSTEMVERILOG SYNTH)
elseif(SIM_TYPE STREQUAL GENERIC)
     ip_compile_definitions(${IP} SYSTEMVERILOG GENERIC)
endif()
```

It can be paired with option 1 as shown to have something configurable.

### Options 3, use new `get_ip_sources()` function

If you want to keep this approach, with the [changes to get_ip_sources()](https://github.com/HEP-SoC/SoCMake/pull/50), it can simplify a little bit what you were doing.

P.S. I know you guys probably know about these things, I wanted to be explicit for later reference and potential users

benoitdenkinger:
Good summary @Risto97. Right now, we actually support all these different ways of differentiating your files. I don't have a strong opinion yet. For me, having a global option (e.g., SYNTH/GENERIC) or a file type (e.g., SYSTEMVERILOG_TB) are both adding 'complexity' to the tool, either way I think both can be understand by users as long as we have a good documentation.

Option 2 is good for distinguish generic vs technology versions, but not really applicable to testbench files. We actually use it for the opentitan prim_cells.

I feel that at the end we either have to support one 'strict' way of doing things and impose a certain standard to how IPs should be written (which is not necessarily a bad thing) or support multiple ways to give more flexibility. I was reading [this](https://veripool.org/papers/TenIPEdits_SNUGBos07_paper.pdf) the other day, old be still valid, so I more in favor have enforcing a certain standard to how code should be written to work properly with socmake.

benoitdenkinger:
Ok, I thought you meant to support SYNTH and GENERIC by SoCMake, but actually you don't, so indeed the complexity does not increase.

Now I'm facing again problems related to that for the FPGA target. I also added SYSTEMVERILOG_TB 'because' we have SYSTEMVERILOG_SIM. Considering these, I agree with your solution (option 1) of letting the user decide which files are included or not (based on some options he is free to define) is probably the most flexible and also simplest (from SoCMake point of view as you said).

About the disadvantage of this option (option 1), I think it is a minor one. Of course, it would be nice to have everything at once, but actually you always test a specific configuration so why should you care about the others. Anyway, we usually end-up with multiple folders for the different file sets/targets (FPGA, SIM, SYNTHESIS) because we often have target specific files and its more clear to keep everything separated.

About using FILE_SETS, indeed it looks similar to what we have but I never tested them. Could we let the user define its own FILE_SETS and have an argument passed to the non-IP targets (e.g., simulators) to specify which file sets to fetch? Or maybe we let the non-IP target even simpler and we add a function to play with the file sets at the IP level. For example, let's say we keep the IP library as simple as possible with only VHDL, VERILOG and SYSTEMVERILOG and all our functions support only these three type of files. The user can still add any property he wants (e.g., ip_sources(IP MY_SET_NAME my_source.sv) and then we have a function like get_ip_from_sets(IP_OUT IP_LIB SETS) that would return a new IP fetching only the specified sets (plus the basic ones VHDL, VERILOG and SYSTEMVERILOG).

Risto97:
> Anyway, we usually end-up with multiple folders for the different file sets/targets (FPGA, SIM, SYNTHESIS) because we often have target specific files and its more clear to keep everything separated.

I think this is the right way to go.

> About using FILE_SETS, indeed it looks similar to what we have but I never tested them. Could we let the user define its own FILE_SETS and have an argument passed to the non-IP targets (e.g., simulators) to specify which file sets to fetch?

Yes, we already kind of have FILE_SETS.
It would be possible to define something like this, but for the release 1.0.0, I would be in favour of leaving only option 1.
Mostly because I am not really sure how to implement the other options nicely and if it would be needed, as you confirmed only 1 file set is usually needed per configuration.
I would prefer adding features after the 1.0.0 release, rather than removing them for backwards compatibility.

> then we have a function like get_ip_from_sets(IP_OUT IP_LIB SETS) that would return a new IP fetching only the specified sets (plus the basic ones VHDL, VERILOG and SYSTEMVERILOG).

The problem with this is that the targets (IP_LIBRARIES) are global once defined, so this will pollute the namespace, and the help menu.

benoitdenkinger:
I agree about keeping 1.0.0 basic and later add features. I'm discussing below just to give some ideas, but I think we should first re-align with the first release and experiment with it, then see if there is important missing features like these FILE_SETS or if we can live without them.

> The problem with this is that the targets (IP_LIBRARIES) are global once defined, so this will pollute the namespace, and the help menu.

Yes we would have one extra target per configuration which may not be easy to read in the help menu (which anyway is always quite long, at least with the version we use now) but you also clearly define the different configuration. For example, let's have you top IP vendor__lib__name__version target then you would get, for example:

- vendor__lib__name__version__fpga
- vendor__lib__name__version__generic
- vendor__lib__name__version__synthesis

So in this case you would get 3 more targets which is not that much. The other option is to do that at the function level (e.g., verilate, xcelium) but this requires more thinking about the proper interface.

PR #96, opened 2024-10-25
copy_rtl function update

benoitdenkinger:
### Changes

1. xcelium function udpated. ELABORATE and UNIQUIFY arguments added.
2. copy_rtl_files function rewritten as a custom target only to force re-copying the file everytime the target is called
3. copy_rtl_files python script updated to error return code for simpler debugging
4. added read_rtl_sources function. It reads a files listings RTL file paths and returns a cmake list. It is meant to be used with copy_rtl_files which creates such a file alongside the copied ones.
5. Minor typo fix, code formatting/cleaning/commenting on tmrg, vivado, and copy_rtl_files functions


Risto97:
What is the reason for the `uniqify` thing?
If I understand the problem is that multiple `IP` blocks contain the same file, but they are physically in different directories.
If this happens, maybe the users should consider making that file a separate IP and link to it, this way redefinition is avoided.

If I remember correctly there was a flag in xcelium or another simulator to allow redefinition, in case user doesn't want to bother, they could use that.
The problem with `uniquify_files_by_basename` is that it reads out whole files and does byte-to-byte comparisons, which might become slow when there are many big files.
So maybe it should be an optional step controlled with a function argument.

Risto97:
To just share an idea I had for "installing" IP blocks, it might do similar to what `copy_rtl_files()` is doing.
There is still quite a bit work to be done to achieve it, so `copy_rtl_files()` is a good option for now.

I do see that there might be a need to move all the sources and generated files to a new directory for "delivery".
It might also not contain the `CMakeLists.txt` but optionally generic file list files as you are doing.

It would be nice to try and use CMake's [`install`](https://cmake.org/cmake/help/latest/command/install.html) phase to "install" IP blocks to new directory.
They can be installed as INTERFACE libraries and later imported with `find_package()` calls.
But most importantly copy all the source files through the `INSTALL_INTERFACE`.

This is something I will have in mind past 1.0.0 release.

benoitdenkinger:
Indeed uniquify_files_by_basename is more of a temporary solution than a deeply needed function. It is linked to copy_rtl_files() as there we loose all the IP definitions and as we do that in different repositories we end-up having multiple times the same files (but with a different path). You could re-instantiate all of the IPs later in a new CMakeLists.txt files, but it's time consuming, so now we get only a single IP once we read the copied files back with a CMakeLists.txt file.

I don't like allowing redefinition because you don't know (unless of reading the log file) what is the final definition you are using and this can easily leads to errors. But I can indeed add a parameter and let the user do what he wants.

Now, you might actually want to have two different versions of an IP, so you will have the same module name but different implementations. You could also have two different IPs having two different modules doing different things but having the same name. I was thinking if we should add the VLNV information to the files to avoid these conflicts. Not manually but transparently to the user.

Right now we are mostly fixing problems quickly but usually not taking advantage of all the cmake built-in functionalities, which feels not right. So I'm in support of any update that tends to better use cmake features (after our deadline ;)), like this install phase.

PR #106, opened 2024-12-02
Refactor simulator functions to allow separate compile and run target…

Risto97:
…s, which allows adding tests with CTest

### DIFF
--- a/SoCMakeConfig.cmake
+++ b/SoCMakeConfig.cmake
@@ -35,8 +35,8 @@ include("${CMAKE_CURRENT_LIST_DIR}/cmake/sim/iverilog/iverilog.cmake")
 # ----- xcelium ------
 include("${CMAKE_CURRENT_LIST_DIR}/cmake/sim/cadence/xcelium.cmake")
 
-# ----- modelsim ------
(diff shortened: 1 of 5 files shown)


Risto97:
With this update, it is possible to create a compilation target without run target for `modelsim`, `iverilog`, `xcelium`, `verilator` using `NO_RUN_TARGET` argument.
Additionally, on subsequent calls to each of the functions, it is possible to create the run target.
Along with `RUN_TARGET_NAME` it is possible to specify the name of the created run target.

This is useful when creating CTest tests, in cases where different test scenarios are for example passed with `$plusargs` CLI arguments.
For example:

```CMake
iverilog(${IP} NO_RUN_TARGET) # Create only compilation target

foreach(test_name hexfile IN ZIP_LISTS test_list hex_list) # Loop through list of cross compiled C tests
    iverilog(${IP}
        RUN_ARGS +firmware=${hexfile}               # Pass path to compiled hexfile to simulation
        RUN_TARGET_NAME run_${test_name}            # Create a target for running the hex file test
        )
    add_dependencies(run_${test_name} ${test_name}) # Make sure hex file is compiled before simulation is invoked
    add_test(NAME run_${test_name}                  # Create a CTest that will run the simulation target
        COMMAND make run_${test_name})
endforeach()

```

PR #107, opened 2024-12-03
s-rec file for jtag booting

benoitdenkinger:
Added:
- gen_srec_files: generates application binary files in srec format

Updated:
- add_tests: added options to control number of processors (NPROC), added a prefix to the test naming (TEST_PREFIX), and added the possibility to call the same test (i.e., same executable), but with different inputs (TESTCASE) passed to an argument (TESTCASE_PARAM).
- xcelium: added SDF_CMD_FILE argument file used for back-annotated simulations


Risto97:
Hey @benoitdenkinger ,

I see you made some changes to Xcelium that overlap with my recent pull request.
We should probably sync up soon to start merging the branches.

I wanted to be able to add tests for simulators other than Verilator.
So far it was not possible as when invoking `xcelium(${IP})` it would create a single target for simulating the IP.

I decided to split the compilation and simulation targets for all the simulators.
So now when you call `xcelium(tb)` it makes 2 targets
2024-12-04T22:48:00,458117020+01:00

If you do the following:

```
xcelium(${IP} NO_RUN_TARGET)

xcelium(${IP}
        RUN_ARGS        +firmware=asd
        RUN_TARGET_NAME run_simple_test)

xcelium(${IP}
        RUN_ARGS        +firmware=bcd
        RUN_TARGET_NAME run_other_test)
```
You will get:
2024-12-04T22:58:32,186605809+01:00

You can pair this with `add_test()` like I described here: https://github.com/HEP-SoC/SoCMake/pull/106#issuecomment-2513169019

I forked a `picorv32` repository to demonstrate this, so you can find example how to use it for `modelsim` (`questa`), `verilator`, `iverilog`, `xcelium`.
https://github.com/Risto97/picorv32_socmake/blob/b3603248d93fa49f6d41f136e0810671f1e1127f/vrf/CMakeLists.txt#L45-L66
I am using callback with `cmake_language()` so it looks a bit weird.

Unrelated, I also used `ExternalProject()` to cross compile the tests. Which is how it should be cross-compiled correctly with CMake:
https://github.com/Risto97/picorv32_socmake/blob/b3603248d93fa49f6d41f136e0810671f1e1127f/firmware/tests/CMakeLists.txt

Let me know if these changes look useful for you, or you might have a better idea of how to implement this.

PR #110, opened 2024-12-12
Allow relative paths for ip_sources() and ip_include_directories(), r…

### LINKED ISSUE #93: ip_sources() allow relative file paths (include directories)
Allow listing relative paths for `ip_sources()` and `ip_include_directories()`.
Files should be still stored as absolute paths, but the absolute paths should be resolved in these function calls.

Risto97:
…elative paths are converted to absolute relative to ${CMAKE_CURRENT_SOURCE_DIR} resolve #93

### DIFF
--- a/cmake/hwip.cmake
+++ b/cmake/hwip.cmake
@@ -2,6 +2,7 @@
 include("${CMAKE_CURRENT_LIST_DIR}/utils/socmake_graph.cmake")
 include("${CMAKE_CURRENT_LIST_DIR}/utils/alias_dereference.cmake")
 include("${CMAKE_CURRENT_LIST_DIR}/utils/safe_get_target_property.cmake")
+include("${CMAKE_CURRENT_LIST_DIR}//utils/file_paths.cmake")
 
 #[[[
 # This function creates an INTERFACE library for a given IP.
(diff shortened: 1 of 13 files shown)


Risto97:
Relative paths are now allowed for `ip_sources()` and `ip_include_directories()` functions.

Functionality matches the CMake [`target_sources()`](https://cmake.org/cmake/help/latest/command/target_sources.html) functionality:

> Changed in version 3.13: Relative source file paths are interpreted as being relative to the current source directory (i.e. [CMAKE_CURRENT_SOURCE_DIR](https://cmake.org/cmake/help/latest/variable/CMAKE_CURRENT_SOURCE_DIR.html#variable:**CMAKE_CURRENT_SOURCE_DIR**)). See policy [CMP0076](https://cmake.org/cmake/help/latest/policy/CMP0076.html#policy:CMP0076).

And [`target_include_directories()](https://cmake.org/cmake/help/latest/command/target_include_directories.html)

> Specified include directories may be absolute paths or relative paths. A relative path will be interpreted as relative to the current source directory (i.e. [CMAKE_CURRENT_SOURCE_DIR](https://cmake.org/cmake/help/latest/variable/CMAKE_CURRENT_SOURCE_DIR.html#variable:CMAKE_CURRENT_SOURCE_DIR)) and converted to an absolute path before storing it in the associated target property. If the path starts with a generator expression, it will always be assumed to be an absolute path (with one exception noted below) and will be used unmodified.

Meaning all relative paths given to `target_sources()` and `target_include_directories()` are converted immediately to absolute paths relative to `CMAKE_CURRENT_SOURCE_DIR` variable.

In short this is now possible
```
add_ip(test)
target_sources(${IP}
    test1.sv
)
```

PR #133, opened 2025-04-03
Develop and master branch synchronization merge

benoitdenkinger:
# Description

The `develop` and `master` branches diverged a few month ago. While the `develop` branch was implementing and improving the features, the `master` branch was going through minor updates to keep the compatibility with existing repositories. This merge synchronize both branches.

# Changes

- Documentation deployment fixed
- SoCMake main features (e.g., HWIP library) tests improved
- Build script for libraries installation: systemC and UVM-systemC
- Simulator functions (e.g., xcelium, verilator, vcs) improved to use tool specific features.
- Simulator function interfaces uniformized
- peakrdl closed-source plugins function removed
- Many examples added to show SoCMake features


Risto97:
Hello @benoitdenkinger.

Well done with the merge, hopefully, we can just continue using develop now.

I have one issue though.
I think you squashed the merge request.

Normally, it's the right way to do it, but in this case, I think we should keep the commit history, as the commits were mostly squashed on develop, so the commit history is already clean.
2025-04-03T18:45:18,390050523+02:00

If you agree, I would propose to do the following.

1. Revert the commit on master, (I made the same mistake here https://github.com/HEP-SoC/SoCMake/commit/c2aa882e1edae8938c4573de1461323792c329c0)
2. Restore this branch and reopen the pull request.
3. In the repository settings and General, disable squash on pull request
2025-04-03T18:49:49,251633894+02:00
4. And now do just a simple merge without squash.

Let me know if you don't have the time to do it, I can do it also.

benoitdenkinger:
Hi @Risto97,

Indeed a non-squash merge request would have been nicer. I'll revert it.

Regarding the develop branch you can continue push on it, but I think it's simpler if we all work on different branches. On our repositories, I'll continue to reference only the master branch as it is the more longer-term branch for referencing commits. In that regard, I think it is also better if we merge often to master to keep everything synchronize and small incremental changes (instead of facing a lot of changes once the develop branch is merge). After the first release, having a develop branch probably make more sense.

benoitdenkinger:
This merge request does not exist anymore in the history and has been replaced by #134

PR #137, opened 2025-04-14
Version conditions for ip_link()

Risto97:
This pull request enables specifying a range of versions that are allowed in `ip_link()` call.
This can be useful to ensure that the correct version of dependent IP is used to avoid issues.

It is now possible to specify the version ranges as following:

```
add_ip(v::l::top)
add_ip(v::l::ip2::3.9.2)

ip_link(v::l::top 
    "v::l::ip2 >= 3.9.2, < 4.0.0"
)
```

Supported relational operators are >, <, >=, <= and ==.
The conditions are separated by a , (comma) symbol.
All the conditions need to be satisfied for the link, otherwise `FATAL_ERROR` is issued.

Let me know what you think of this feature.

### DIFF
--- a/cmake/hwip.cmake
+++ b/cmake/hwip.cmake
@@ -418,6 +418,58 @@ function(check_languages LANGUAGE)
     endif()
 endfunction()
 
+
+function(__compare_version RESULT COMPARE_LHS RELATION COMPARE_RHS)
+
+    if(RELATION STREQUAL ">=")
+        set(RELATION VERSION_GREATER_EQUAL)
+    elseif(RELATION STREQUAL ">")
+        set(RELATION VERSION_GREATER)
+    elseif(RELATION STREQUAL "<=")
+        set(RELATION VERSION_LESS_EQUAL)
+    elseif(RELATION STREQUAL "<")
+        set(RELATION VERSION_LESS)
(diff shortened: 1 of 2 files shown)


benoitdenkinger (commented):
Nice feature. From the code it seems that any number of conditions can be given (even without the comma), for example:

ip_link(v::l::top "v::l::ip2 == 2.9.2, == 3.2.3, => 1.2.5")

Should we add some restrictions or this is unnecessary as adding more conditions is either meaningless or may create an unsatisfied requirement which is up to the user to resolve?

In general, should we add more tests to check dummy/wrong formats are properly detected?

Risto97:
Good catch, the regex would just ignore the comma.
It might be tricky to write regex to catch the comma, I will give it a try.
Yes, certainly this needs have a test.

I got moderately inspired by cargo dependency syntax https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#version-requirement-syntax

You are right, more than 2 dependencies probably do not make sense, unless there is a possibility for OR condition.
But this would be going too far maybe.
The safest would be to constrain it to 1 or 2 requirements, and more than that report an error.

benoitdenkinger (approved):
Nice add, it's good to go for me

PR #139, opened 2025-05-05
Cocotb update

benoitdenkinger:
Updated the cocotb function to a more generic structure and formatted it to the standard interface of other simulators.

The cocotb is not a simulator but a python library that can be used to create testbenches. Therefore, the cocotb function is a wrapper around multiple simulators. For now only the icarus verilog, xcelium and verilator simulators are supported and it misses vcs and modelsim/questasim (they are supported by cocotb and shouldn't be hard to support)


cmake/sim/cocotb/cocotb.cmake
```diff
@@ -0,0 +1,287 @@
...
+    # find_package(Python3 COMPONENTS Interpreter Development)
+
+    alias_dereference(IP_LIB ${IP_LIB})
+
+    if(NOT ARG_TOP_MODULE)
+        get_target_property(ARG_TOP_MODULE ${IP_LIB} IP_NAME)
+    endif()
+
+    if(ARG_PYTHONPATH)
+        # Column separated paths to python files/modules are needed
```
Risto97, line 55: Why do you need to set the PYTHONPATH manually?
Should it not be dealt by the venv?

cmake/sim/cocotb/cocotb.cmake
```diff
@@ -0,0 +1,287 @@
...
+        COMMAND ${COCOTB_CONFIG_EXECUTABLE} --lib-dir
+    )
+    # Remove the line feed of the variable
+    string(STRIP ${COCOTB_PY_DIR} COCOTB_PY_DIR)
+    string(STRIP ${COCOTB_LIB_DIR} COCOTB_LIB_DIR)
+    set(COCOTB_SHARE_DIR ${COCOTB_PY_DIR}/cocotb/share)
+    # First get all Python files from cocotb
+    file(GLOB_RECURSE COCOTB_PY_DEPS ${COCOTB_PY_DIR}/*.py)
+    # Get all files in the cocotb library directory
+    file(GLOB COCOTB_LIB_DEPS ${COCOTB_LIB_DIR}/*.so)
```
Risto97, line 84: Does it make sense to list them explicitly without globing?

cmake/sim/cocotb/cocotb.cmake
```diff
@@ -0,0 +1,287 @@
...
+        set(sim_build_dep ${IP_LIB}_iverilog)
+    elseif(${ARG_SIM} STREQUAL verilator)
+        message(DEBUG "COCOTB: Using Verilator simulator")
+
+        verilator(${IP_LIB}
+            NO_RUN_TARGET
+            TOP_MODULE ${ARG_TOP_MODULE}
+            DIRECTORY ${cocotb_sim_build}
+            PREFIX Vtop
+            VERILATOR_ARGS --Wno-fatal -DCOCOTB_SIM=1 --vpi --public-flat-rw
```
Risto97, line 140: Does cocotb require --Wno-fatal to be set?

Risto97: Also --public-flat-rw degrades the performance of verilated model massively, it would be better to selectively declare signals public.
But this is maybe a cocotb limitation?

cmake/sim/cocotb/cocotb.cmake
```diff
@@ -0,0 +1,287 @@
...
+        )
+
+        add_executable(cocotb_verilator
```
Risto97, line 143: Maybe the executable target should have the ${IP_LIB} in the name, to avoid the name clashes in case you call the cocotb() function twice in the project

cmake/sim/cocotb/cocotb.cmake
```diff
@@ -0,0 +1,287 @@
...
+            set(ARG_GUI GUI)
+        endif()
+
+        xcelium(
+            ${IP_LIB}
+            NO_RUN_TARGET
+            ${ARG_GUI}
+            TOP_MODULE ${ARG_TOP_MODULE}
+            OUTDIR ${cocotb_sim_build}
+            ELABORATE_ARGS -access +rwc -timescale 1ns/1ps -loadvpi ${COCOTB_VPI_PATH}:vlog_startup_routines_bootstrap -loadvhpi ${COCOTB_VHPI_PATH}:cocotbvhpi_entry_point
```
Risto97, line 179: Similar to verilator --public-flat-rw, there is -access +rwc, it should be only passed we debug the design.
Unless its a cocotb requirement.
Probably the same goes for -timescale argument

cmake/sim/iverilog/iverilog.cmake
```diff
@@ -27,6 +37,16 @@ function(iverilog IP_LIB)
     # Include the hardware IP management cmake functions
     include("${CMAKE_CURRENT_FUNCTION_LIST_DIR}/../../hwip.cmake")
 
+    # Check the executables are available
+    find_program(IVERILOG_EXECUTABLE iverilog)
+    if(NOT IVERILOG_EXECUTABLE)
+        message(FATAL_ERROR "iverilog not found in PATH. Please install icarus verilog or add it to your PATH.")
```
Risto97, line 43: I am not against throwing an error during configure time if the executable is not found.
Another option is to let the configure phase go without errors, and let Make fail.

I think the error during configure is probably a safer bet, and more true to how CMake works.
It also makes the CMake project, when configured, more stable, as it does not depend on the environment variables and the path.
If we decide on this option, we should apply the same principle to all simulators

Risto97:
Hello @benoitdenkinger,
Great work, I like that you managed to reuse the existing simulator support functions instead of letting cocotb run the compilation.
I added some comments in the pull request.

Just to brainstorm, I wonder if in the future we can have something like this:

```
add_ip(tb)
...

ip_link(tb
    SoCMake::cocotb_support # Adds the cocotb VPI libraries
)

xcelium(${IP} NO_RUN_TARGET)

cocotb(${IP})
```

Instead of
```
......
cocotb(${IP} SIM xcelium)
```

Additionally, can you please add an example of a simple cocotb testbench, it should have an option to choosethe  simulator through cmake configure options like here:
https://github.com/HEP-SoC/SoCMake/blob/master/examples/simple_verilog/CMakeLists.txt#L6
I would like to test it to see how it works.

Risto97 (changes requested):

cmake/sim/cocotb/cocotb.cmake
benoitdenkinger: I re-used the same approach that is implemented in the cocotb makefile. This is also simpler to support future possible changes in file

cmake/sim/cocotb/cocotb.cmake
benoitdenkinger: Yes you are right

cmake/sim/cocotb/cocotb.cmake
benoitdenkinger: -access +rwc is indeed required for cocotb, not sure about the timescale

cmake/sim/iverilog/iverilog.cmake
benoitdenkinger: I'm not in favor of one approach yet. I like to detect the problem as soon as possible, but it also prevent the configuration phase to finish, as you mentioned.

Sometimes it's nice to see the invoked command even if you cannot run it, but it is mostly useful while developing/debugging. On the other hand, if configuration takes a bit of time, you may want to detect problem as soon as possible

cmake/sim/cocotb/cocotb.cmake
Risto97: Ok, makes sense, well the danger is that these files update, but CMake is not reconfigured, as the globing is done at configure time.
But I would say the risk is pretty low

cmake/build_scripts/verilator/verilator_build.cmake
```diff
@@ -0,0 +1,62 @@
...
+            COMMAND_ECHO STDOUT
+        )
+
+        execute_process(COMMAND ${CMAKE_COMMAND}
+            --build ${CMAKE_BINARY_DIR}/verilator-build
+            --parallel 4
+        )
+    endif()
+
+    find_package(verilator 5.012 REQUIRED
```
Risto97, line 54: Version is hard coded

cmake/build_scripts/verilator/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
+cmake_minimum_required(VERSION 3.25)
+project(verilator_build)
+
+include("../../utils/option.cmake")
+
+set(VERILATOR_VERSIONS 5.012)
```
Risto97, line 6: Also here, the version

Risto97:
Just have a look also at my comment maybe you missed https://github.com/HEP-SoC/SoCMake/pull/139#issuecomment-2855914594

An example would be good so I can test it, and people in general can give it a try

cmake/sim/cocotb/cocotb.cmake
benoitdenkinger: I was not aware of the potential issue related to critical signals like clocks, good to keep in mind

cmake/build_scripts/verilator/verilator_build.cmake
benoitdenkinger: Yes, I'm doing some tests, I should set it to draft for now

benoitdenkinger:
I forgot to post my answer to [#139](https://github.com/HEP-SoC/SoCMake/pull/139#issuecomment-2855914594)...

Yes, I'll add an example. Regarding your proposal for future improvement, it might be feasible. Now, you can already avoid setting the simulator and it will try to find one among the supported simulators (in a predefined order). The only problem to solve for your proposal is to retrieve the simulation command to run.

cmake/build_scripts/verilator/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ project(verilator_build)
 
 include("../../utils/option.cmake")
 
-set(VERILATOR_VERSIONS 5.012)
+set(VERILATOR_VERSIONS 5.012 5.014 5.016 5.018 5.020 5.036)
```
Risto97, line 6: Will it fail for older versions than 5.012?
Should we just let this be a string and if the version is incorrect, the git checkout will fail.
Not the most elegant, but we wouldn't need to update the enumeration every time new version is out

Risto97:
I managed to get it running, the iverilog and verilator work. I have issues with Xcelium, but this might be something to do with version, or who knows. I will investigate when I have time.

Well done!

Risto97:
Do you think it would make sense to package cocotb python files into an IP libraries?

The example would look something like this:

```
add_ip(adder_test)
ip_sources(${IP} COCOTB_PY
    ./simple_cocotb_example.py
)

add_subdirectory(adder)

ip_link(adder_test adder)

cocotb(adder_test
    # TOP_MODULE adder # Is it needed? It would automatically compile each IP library that is linked to test
    # COCOTB_MODULE ${PROJECT_NAME} # Can be inferred from the adder_test
    SIM ${SIMULATOR}
    OUTDIR ${PROJECT_BINARY_DIR}
    # PYTHONPATH ${CMAKE_CURRENT_LIST_DIR} # This would not be necessary as it would be made internally by getting the parent directories of all the COCOTB_PY files
)
```

benoitdenkinger:
> The ExternalProject was there to solve an issue with verilate() function, and problem of calling it directly in configure phase

Yes, I realized that yesterday while changing the function. I will revert it.

> I have issues with Xcelium

I set it up but didn't test it yet, I'll check if it work for me

> Do you think it would make sense to package cocotb python files into an IP libraries?

This could indeed be an option. For now I would prefer to keep this version. I don't want to add a new set of files too fast

benoitdenkinger:
The cocotb example with xcelium is working for me

Risto97:
CMake has `VERSION_EQUAL` for comparing versions in a `if()` clause, its a bit more robust than doing STREQUAL.
I also added a check for the `ARG_INSTALL_DIR` before compiling.

There is also an issue with Verilator and cocotb.
Here are the steps to reproduce:

```
cmake -DSIMULATOR=verilator ../
make run_adder_cocotb_simple_cocotb_example
touch ../adder/adder.v
make run_adder_cocotb_simple_cocotb_example
```
What I get is
```
[  9%] Performing build step for 'adder_verilate'
[  6%] Generating Vtop.cpp, Vtop___024root__DepSet_h84412442__0.cpp, Vtop___024root__DepSet_heccd7ead__0.cpp, Vtop___024root__Slow.cpp, Vtop___024root__DepSet_h84412442__0__Slow.cpp, Vtop___024root__DepSet_heccd7ead__0__Slow.cpp, Vtop__Dpi.cpp, Vtop__Syms.cpp, Vtop.cmake
%Error: $VERILATOR_ROOT needs to be in environment
make[6]: *** [CMakeFiles/adder.dir/build.make:77: Vtop.cpp] Error 1
```

Risto97:
> The cocotb example with xcelium is working for me

Yes, got Xcelium to work.
Cocotb doesn't seem to work well with pyenv, also on Centos7 it doesn't work, but nothing we can do in SoCMake about that.

I think it will be fine to close this pull request once Verilator issue is resolved.
But just to understand, how much effort would it be to add VCS, Questa, Vivado simulator and maybe GHDL?
Do you plan on adding those also?

benoitdenkinger:
> %Error: $VERILATOR_ROOT needs to be in environment

Fixed.

> But just to understand, how much effort would it be to add VCS, Questa, Vivado simulator and maybe GHDL?
Do you plan on adding those also?

I added VCS. I tried Questa and could make it work in 30 minutes so I quit, but I think it shouldn't be to hard. GHDL is for VHDL and there is no need on our side (so very low priority). Vivado simulator is not officially supported by cocotb so I didn't even try.

Now we have the main simulators (except questa maybe) and I have higher priorities so I won't spend more time on it now.

Risto97:
Ok, I could run all the simulators.
It looks ready to merge with me. Great work!

> I added VCS. I tried Questa and could make it work in 30 minutes so I quit, but I think it shouldn't be to hard. GHDL is for VHDL and there is no need on our side (so very low priority). Vivado simulator is not officially supported by cocotb so I didn't even try.
> Now we have the main simulators (except questa maybe) and I have higher priorities so I won't spend more time on it now.

Yes, that makes sense.

Risto97 (dismissed):

PR #143, opened 2025-05-20
UVM-SystemC build script fixed - fix #149

benoitdenkinger:
The UVM-SystemC `--with-arch-suffix`argument needs either to be empty, for `lib`, or 64, for `lib64` to find the SystemC library. A small piece of code is checking the value of CMAKE_INSTALL_LIBDIR to automatically set the correct value.

### DIFF
--- a/cmake/build_scripts/uvm-systemc/uvm-systemc_build.cmake
+++ b/cmake/build_scripts/uvm-systemc/uvm-systemc_build.cmake
@@ -22,8 +22,8 @@ function(uvm_systemc_build)
 
     # TODO ARG_VERSION cannot be used as its not following major.minor.patch
     find_package(UVM-SystemC CONFIG
-        HINTS ${ARG_INSTALL_DIR}
-        )
+        HINTS ${UVM_SYSTEMC_HOME} $ENV{UVM_SYSTEMC_HOME} ${ARG_INSTALL_DIR}
+    )
     get_target_property(SYSTEMC_INC_DIR SystemC::systemc INTERFACE_INCLUDE_DIRECTORIES)
     set(SYSTEMC_HOME "${SYSTEMC_INC_DIR}/../")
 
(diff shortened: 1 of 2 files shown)


Risto97:
This things are tricky to get right as they depend on OS, CPU architectures, etc...
A lot of testing on different platforms is needed to confirm it.

But I tested on following:

ArchLinux with GCC 15.1.1 works
Centos7:
* GCC 4.8.5 (doesn't work, but not an issue related to your change, but uvm-systemc needs newer GCC, maybe -std17 issue)
* GCC 10.2.1 is not working, but now the issue is the lib, lib64

The installed lib and lib64 looks like this:

```
lib
└── cmake
    └── UVM-SystemC
        ├── UVM-SystemCConfig.cmake
        ├── UVM-SystemCConfigVersion.cmake
        └── UVM-SystemCTargets.cmake
lib64
├── libuvm-systemc-1.0-beta6.so
├── libuvm-systemc.a
├── libuvm-systemc.la
├── libuvm-systemc.so -> libuvm-systemc-1.0-beta6.so
└── pkgconfig
    └── uvm-systemc.pc
```

So when trying to run `make` after the deps are built, I get this:

```
make[2]: *** No rule to make target `/tools/shared/work/SoCMake/examples/uvm-systemc/deps/_deps/uvm-systemc/lib/libuvm-systemc.so', needed by `systemc_example'.  Stop.
make[1]: *** [CMakeFiles/systemc_example.dir/all] Error 2
make: *** [all] Error 2
```

Anyways, I tried to fix it, by making sure the both cmake package files and libraries go into the same directory on the platform.
So for me now Archlinux and Centos7 (GCC 10.2.1) passes.
Which OS and GCC are you testing this on?

Risto97 (approved):

Risto97:
BTW. I only tried it on:
[uvm-system-example/](https://github.com/HEP-SoC/SoCMake/tree/master/examples/uvm-systemc)

PR #160, opened 2026-01-07
Excluded ips

benoitdenkinger:
Added EXCLUDED_IPS keyword argument to fetch IPs without certain dependencies with a simplified function to replace flatter_graph.


benoitdenkinger:
When moving to implementation, dividing a design into hierarchical blocks is often needed. In this case, it could be nice to get all the sources from one IP excluding the sources of some lower levels hierarchical blocks in order to implement only the needed part.

Could you (re)explain me the reason for the need of the topological sorting?

I didn't see a way to do that easily with the current solution, so I wanted to try doing it another way to test if such a feature works and is interesting. What I did is to rely on the INTERFACE_LINK_LIBRARIES property to traverse the IPs and get the source files. This way, I can easily stopped at a IP/hierarchical level provided by the EXCLUDED_IPS parameter. It's probably not working for everything but this is for now only to test the EXCLUDED_IPS feature (actually I opened the PR mostly to test the gitlab-ci-approved labeling to trigger the gitlab CI from a fork PR).

benoitdenkinger:
> But what if you have something above, if you exclude B should D be excluded or not? Because C also depends on it.

If you exclude B, D will remain as long as it is linked by another IP. If not, it is excluded when B is.

> But why don't you just use sources of IP1?
You can do that by using NO_DEPS argument in all get_ip_...() Functions, like so:

For very simple design yes, but for more complex ones it is often not good enough.

> Note that in regular CMake there a concept of PRIVATE, PUBLIC, INTERFACE, but we don't have that in SoCMake as it's not very clear how to do it for HDL languages.

Yes I know, but sometimes we do link other kind of targets (e.g., C or C++ ones) and I was not sure if this change could break this. It shouldn't, as long as the functions of HWIP are used only for HDL sources. But for example the function get_ip_property is quite generic and I don't know if it was ever used to retrieve more standard cmake properties and where getting only the INTERFACE targets is not enough. We can also declare such use cases (if any) are not the intended way of using the HWIP functions.

> I was worried that this would break the tests

Indeed, the test you mentioned (ip_link1) failed at the beginning because I was appending instead of prepending, but now it works.

PR #161, opened 2026-01-08
Gitlab trigger

benoitdenkinger:
Added workflow job to trigger gitlab CI through POST request to giltab REST API


benoitdenkinger:
Re-applied in a different PR (https://github.com/HEP-SoC/SoCMake/pull/162) after history problems

PR #169, opened 2026-01-19
Change Modelsim function to questasim to be up-to-date with the tool name/version

mtravaillard:
The dpi-c program was forced to be compiled into a 32-bit library when using Modelsim, and Modelsim is expecting a 64-bit library.

When calling Modelsim in a CMakeList file, we can specify to run a 32bit version, which will had those flags and tell Modelsim to run in 32bit.

It's not done in the [example](https://github.com/HEP-SoC/SoCMake/blob/571d08365b55fea0f64fa2c2facecdfce58088ee/examples/dpi-c/CMakeLists.txt#L31) so we don't need to add them.

Edit: To make it easier to understand if you look at this PR in the future, it was initially done to fix a problem with the dpi using modelsim. As the discussion under can show, we ended up changing the function to be up-to-date with how the tool evolved, as "modelsim" is discontinued and have been replaced by "questasim"


Risto97:
Hello @mtravaillard ,

This can be a bit confusing, but there is a difference between the free Modelsim version and commercial Questasim.
They actually share the same CLI interface, but there are quite a lot of differences in the tools itself.

One of the differences is that the free Modelsim is 32bit and can link only 32bit binaries.
Questasim is 64bit (possibly can link 32bit binaries?).

So in this particular example, if someone uses Questa you should use `-DSIMULATOR=questa` to compile 64bit DPI library.
It would be nice if there was some way to identify if `vsim/vlog` are Modelsim or Questasim and add flags based on that?

P.S.
I might be wrong, and maybe the newer versions of Modelsim are 64bit?

mtravaillard:
Hello @Risto97

I understand why it can be confusing...

I don't have a free Modelsim version to try, maybe doing a `vsim/vlog --version` or something similar can help to tell which one of the tool is being used and if it's running on 32 or 64 bit ? We could then grab the information from this and set the argument for the 32 bit version if needed.

Otherwise, I can add Questa to the list of simulator and depending on the one being used, adding the 32bit argument when calling the tool in the CMakeList file. But this would keep the confusion and it would have to be explained in the documentation or in the `help_options.txt` generated file...

PS: From what I saw on internet, all the new version are called Questa and there is no Modelsim anymore (since 2020 apparently). The free version, Questa Starter Edition is still in 32bit. I think that's why it's really confusing :laughing: 

Edit: It looks like when they switched from Modelsim to Questasim, they stopped working with 32-bit version. So Questa Free version is in 64 bit but Modelsim free version, that can't be downloaded anymore from their website, is 32bit.
So should we really consider supporting Free Modelsim version if they stopped doing them 4 years ago ?

benoitdenkinger:
If modelsim is always 32 bits and questasim always 64 bits, instead of removing the code we just need to distinguish modelsim and questasim in the simulator options and use the correct one depending on the setup. In the case of our gitlab CI runners, we will use questasim, but if someone use modelsim it still works.

It is true that modelsim is slowly disappearing but it doesn't cost much to support it in this case.

Risto97:
> Edit: It looks like when they switched from Modelsim to Questasim, they stopped working with 32-bit version. So Questa Free version is in 64 bit but Modelsim free version, that can't be downloaded anymore from their website, is 32bit.
> So should we really consider supporting Free Modelsim version if they stopped doing them 4 years ago ?

That would be great, I would say as long as there is a free simulator that is available, we need to support it.
But we don't need to support the older version of that simulator, so if free questasim replaces old modelsim, we can just write that old modelsim version is not supported.
With that said, I think the only issue is basically with compiling the DPI-C extensions, most other things are mostly compatible between the two.

Could you also send me the link where do you get the free version of questasim, because different FPGA vendors provide their own modelsim distribution to download

mtravaillard:
Hi !
I downloaded it from the [Altera website](https://www.altera.com/downloads/simulation-tools/questa-fpgas-standard-edition-software-version-25-1).
I was able to install it on my local machine, I haven't been able to get a free license when i tried, it was a captcha problem and i haven't tried since.
But when I did `vsim -version` it told me that it was the 64 bit version and when i launched script it was showing 64-bit tools but it wasn't working as i don't have a license

Risto97:
Hello @mtravaillard ,

Thank you, I tried it and it does work.
It is indeed more complicated to install than modelsim as the account needs to be created and license files setup.

I would say as long as there is the 64bit free version, we can drop the old modelsim support, so I approve this pull request.
@benoitdenkinger do you agree, or you think there is still value trying to support modelsim?

The question is now should we rename every instance of modelsim to questasim?

benoitdenkinger:
modelsim is meant to disappear and I think now is a good time to stop the support/reference, especially if a free version of questasim is available (and in 64b). In this case, we should also rename everything into questasim indeed. @mtravaillard you can add this to this PR.

Risto97:
Ok I agree,
Can we just deprecate the `modelsim()` function for now instead of removing it immediately, as I still have builds that use modelsim.

Here is how to easily deprecate a function https://youtu.be/bsXLMQ6WgIk?t=741

mtravaillard:
Hi @Risto97 @benoitdenkinger,

For now, I added a questasim function (that is exactly the same as modelsim).
I added a deprecation message if someone is calling modelsim function but I haven't changed the behavior.
I have also updated some examples, to work with new questasim functions.

For now, I kept the support with modelsim, as like a "transition stage".
Should I completely remove it and just send a deprecation message when calling modelsim ? Should I keep it in the examples (it might be confusing if it's still here...) ?

Risto97:
Hello @mtravaillard ,

What I meant was something in line of:

Make modelsim a macro() and add deprecation warning. 
In the macro call new questasim function while forwarding all arguments. 
You can use the example I sent you from the video and implement it exactly as shown there.

This way we have no maintenance effort for maintaining the modelsim function, as it has almost no code.

From examples, in my opinion you can delete all references to modelsim at this point.
Also that parallel example for modelsim you can also remove, it should be more of a test than an example.

SoCMakeConfig.cmake
```diff
@@ -40,7 +40,7 @@ include("${CMAKE_CURRENT_LIST_DIR}/cmake/sim/cadence/xcelium.cmake")
 include("${CMAKE_CURRENT_LIST_DIR}/cmake/sim/synopsys/vcs.cmake")
 
 # ----- siemens ------
-include("${CMAKE_CURRENT_LIST_DIR}/cmake/sim/siemens/modelsim.cmake")
```
Risto97, line 43: We need to keep both lines, as the modelsim function would not be callable if we remove it from here

Risto97 (changes requested):
Just this small change and It looks good to me

SoCMakeConfig.cmake
mtravaillard: it's working without it for me.
If you think it's cleaner this way, i will make a `modelsim.cmake` file with only the macro inside, that will call questasim

Risto97: Ah no, sorry I didnt see well.
What you did is correct, adding the modelsim macro to questasim.cmake file.

benoitdenkinger (approved):

Risto97 (approved):

PR #170, opened 2026-01-21
Add custom target to run some examples in a single command

mtravaillard:
By using `add_custom_target()` cmake function, we are able to create a target `run_executable`.
CMake will generate the right Makefiles, allowing to use this new target, that will do everything needed to compile the executable and then run it.

Before adding this custom target, we were only able to compile the executable correctly and the executable had to be launched manually.

This would be helpful to improve the CI.

### DIFF
--- a/examples/dpi-c/CMakeLists.txt
+++ b/examples/dpi-c/CMakeLists.txt
@@ -55,6 +55,8 @@ if(SIMULATOR STREQUAL "verilator")
     verilator(${IP})
     add_executable(verilator_tb EXCLUDE_FROM_ALL Vtb__main.cpp )
     target_link_libraries(verilator_tb tb__vlt)
+    add_custom_target(run_tb_verilator
+        COMMAND verilator_tb)
 endif()
 
 help()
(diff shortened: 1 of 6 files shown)


mtravaillard:
Hi @Risto97,

I'm asking you there because, if I have to change something, I will include it with this PR has it's kind of relevant.
For the `simple_mixed_language_sc_vlog` example, modelsim is referenced as mgc in the [simulator option](https://github.com/HEP-SoC/SoCMake/blob/571d08365b55fea0f64fa2c2facecdfce58088ee/examples/simple_mixed_language_sc_vlog/CMakeLists.txt#L58).

I was wondering if there is a reason or if it haven't been updated when the file has been modified over time ?

It would be helpful if I could change it to `modelsim`, this way I could easily add this test in our internal CI and keep it clear.
Otherwise, if it has to be kept like this, I will find a way don't worry.

Risto97:
Hello @mtravaillard ,

Sorry for that, I was just not being consistent.
Yes feel free to rename it to `modelsim`.

Whenever we can be more consistent is good in general

Issue #140, opened 2025-05-12
Use C++ package managers instead of build scripts


Risto97:
Currently, we have custom CMake functions for building libraries and tools: https://github.com/HEP-SoC/SoCMake/tree/master/cmake/build_scripts

This can work quite well, but it's inefficient for us to support this.
Since there are already projects that deal with this issue, we should consider just using those.

Here is a list of C++ package managers that I tried:

1. [Hunter](https://github.com/cpp-pm/hunter)

* Implemented on CMake
* Integrates into CMake project
* Looks similar to current build_scripts

To add package, simply call in the CMake project
```
hunter_add_package(ZLIB)
```
More examples here [https://github.com/cpp-pm/gate](https://github.com/cpp-pm/gate)

2. [vcpkg](https://github.com/microsoft/vcpkg)

* Implemented in CMake
* Developed by Microsoft
* Can be used in CLI mode or in manifest mode, where it requires separate manifest json file
* Only manifest mode can do versioning of packages
* The CMake build scripts are unaware of vcpkg, need to pass `-DCMAKE_TOOLCHAIN_FILE` when configuring

3. [conan](https://github.com/conan-io/conan)

* Implemented in Python, which makes it a bit tricky to use
* Has a conan center repository of packages https://conan.io/center
* In theory already has Verilator and SystemC packages, but they seem outdated.
* Should have good CMake integration

#139 
@benoitdenkinger

Risto97:
Hello @adrianf0, Welcome to the project.

@benoitdenkinger  @adrianf0 
I have never used Nix, but I am aware of it.
I think its a great way of getting the dependencies, and with what I listed in the first comment is one of the possibilities for the users to get the dependencies.

The initial build scripts came because we did not have possibilities to install any packages without root permissions, so the only option was to build it from source.
Nix was probably option also at that time as it could been run as user without root permissions, but I didn't try it.

With that said, it is probably a good idea to leave this to the users to decide on what they would use.
What we could do is maybe provide a page in the documentation or a blog post on how to use these package managers with SoCMake.
That would mean we would delete the current build scripts to reduce the maintenance effort, which I am in favor of.

And as Adrian said focus on managing project sources, EDA tools and HW-SW co-simulation.

Risto97:
BTW.

If we want to have a more tighter integration with Nix or any other package manager, we can consider something like https://github.com/Risto97/uvtarget does for Python dependencies, where Python packages are installed at configure time and virtual environment is created.

[...]

Issue #119, opened 2025-02-12
Move to internal CERN repository


Risto97:
The following files we agreed to move to CERN private repository:

- [ ] firmware/linker_script
- [ ] add_test.cmake (Add a simpler one as an example, maybe from [picorv32_socmake](https://github.com/HEP-SoC/picorv32_socmake/blob/main/firmware/tests/CMakeLists.txt)
- [ ] peakrdl_docusaurus.cmake (move it for now until its open source)
- [ ] peakrdl_socgen.cmake (I don't remember what we said, either move it or not, I don't mind)
- [ ] peakrdl_ipblocksvg.cmake
- [ ] peakrdl_regblock_wrap.cmake
- [ ] hwif_sed.sh
- [ ] tmake/
- [ ] sv2v.cmake (we will use the one on develop, the master one will be removed)
- [ ] sed_wor.cmake
- [ ] uniqify_files_by_basename.cmake
- [ ] verilate_xml.cmake, scripts/
- [ ] verisc/

PR #174, opened 2026-01-24
Bootstrap scripts for SoCMake

### LINKED ISSUE #173: SoCMake simplified installation
Currently the recommended way of getting SoCMake into a project is either with CPM.cmake or just plain FetchContent.
Then there are also options of fetching it with package managers like Nix as discussed in #140 , however Nix is not available in every environment.

I would like to have something lighter in the build scripts, ideally the `CMakeLists.txt` should look like this:

```cmake
cmake_minimum_required(VERSION 3.30)
project(find_socmake NONE)

find_package(socmake 0.3.1 EXACT REQUIRED)

# Now we can use SoCMake
add_ip(tb)
```

In order for this to work SoCMake needs to be installed on the machine, but we typically don't want to install a fixed version of SoCMake system wide, as it makes it difficult to change it per project.

I will discuss the solution in a pull request

Risto97:
Here is the solution to #173 .

A bootstrapping script is added to make fetching SoCMake easier.
The bootstrapping script allows for SoCMake to be fetched in the project like so:

```cmake
cmake_minimum_required(VERSION 3.30)
project(find_socmake NONE)

# set(SOCMAKE_GIT_TAG b4f050d)
# set(SOCMAKE_GIT_URL ssh://corporate_git_repo/project/SoCMake.git)
find_package(socmake 0.3.1 EXACT REQUIRED)

add_ip(tb)
```

This form should work both in cases when bootstrapping script is used, or if a system-wide SoCMake installation (Nixpkg) is used (as proposed by @adrianf0 #140 ) .

User needs to run once the following command:
```bash
curl -fsSL https://raw.githubusercontent.com/risto97/socmake/bootstrap/bootstrap/bootstrap.sh | sh
```

This will install 2 files in `~/.local/lib/cmake/socmake`: `socmake-config.cmake` and ` socmake-config-version.cmake`.

These 2 files make a CMake package that the `find_package()` call can find, and its placed in a standard location where CMake searches for packages, without having to add any environment variables.
Also these files are version agnostic, so they should not change with SoCMake updates, what they do is simply forward the `find_package()` call to `FetchContent` instead.
The real repository is cloned into the `FETCHCONTENT_BASE_DIR`, and included into the project.

-----------------

There are 2 variables that influence the bootstrap CMake package.

* `SOCMAKE_GIT_TAG` can be used to specify a commit hash or a branch name, while passing version to `find_package()` can be used only if the version is in SEMVER format, due to how `find_package()` works.
* `SOCMAKE_GIT_URL` in order to specify an alternative repository to download.

In case `find_package(socmake)` is called without version argument, the `develop` branch will be cloned.

----------------

I propose releasing these bootstrap scripts under `MIT` license, so that users can modify the default git url to match their git server.

@mtravaillard , @benoitdenkinger Let me know what you think of this solution.
Should we keep in the readme the `FetchContent` instructions, or just add a link to the website documentation on how to use `CPM`, `FetchContent` or `Nix` (in the future) as additional ways of fetching it?

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,20 +38,22 @@ Getting started
 SoCMake is lightweight and has minimal dependencies.
 The only mandatory dependencies are `CMake>=3.25.0` and `make` ([Install Dependencies](https://hep-soc.github.io/SoCMake/docs/getting_started))
 
-There is no need to install SoCMake on your system, it is possible to fetch it in your `CMakeLists.txt`
+### One-Time Bootstrap Installation (Recommended)
 
-Create a file called `CMakeLists.txt`
+Install the SoCMake bootstrap files once on your system:
+**TODO Change the repo and branch name after merging**
+```bash
+curl -fsSL https://raw.githubusercontent.com/risto97/socmake/bootstrap/bootstrap/bootstrap.sh | sh
+```
+
+This installs lightweight bootstrap files to `~/.local/lib/cmake/socmake` that automatically fetch the correct SoCMake version for each project.
+In order to use it create a file called `CMakeLists.txt`
 
 ```CMake
-# Bootstrap SoCMake into CMake project
-include(FetchContent)
-FetchContent_Declare(SoCMake
-    GIT_REPOSITORY "https://github.com/HEP-SoC/SoCMake.git"
-    GIT_TAG develop)
-FetchContent_MakeAvailable(SoCMake)
-
 cmake_minimum_required(VERSION 3.25) # CMake minimum required version
 project(adder NONE)                  # Name of CMake project
+
+find_package(socmake REQUIRED)
 
 # Create an IP block called adder
 add_ip(adder)  
--- a/bootstrap/bootstrap.sh
+++ b/bootstrap/bootstrap.sh
@@ -1,7 +1,9 @@
 #!/bin/bash
 set -e
 
-SOCMAKE_VERSION="${SOCMAKE_VERSION:-develop}"
+# TODO We should fix this to a tag that we know contains these scripts
+# These scripts should not change in the future, as there is no version information
+SOCMAKE_VERSION="${SOCMAKE_VERSION:-bootstrap}"
 INSTALL_DIR="${HOME}/.local/lib/cmake/socmake"
 # TODO change this once merged to main repo
 REPO_URL="${SOCMAKE_REPO_URL:-https://raw.githubusercontent.com/risto97/socmake}"
--- a/bootstrap/bootstrap.sh
+++ b/bootstrap/bootstrap.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+set -e
+
+SOCMAKE_VERSION="${SOCMAKE_VERSION:-develop}"
+INSTALL_DIR="${HOME}/.local/lib/cmake/socmake"
+# TODO change this once merged to main repo
+REPO_URL="${SOCMAKE_REPO_URL:-https://raw.githubusercontent.com/risto97/socmake}"
+
+echo "Installing SoCmake bootstrap files..."
+echo "  Git tag: ${SOCMAKE_VERSION}"
+echo "  Install directory: ${INSTALL_DIR}"
+echo "  Repository: ${REPO_URL}"
+
+# Create installation directory
+mkdir -p "${INSTALL_DIR}"
+
+# Download bootstrap files
+echo "Downloading bootstrap files..."
+curl -fsSL "${REPO_URL}/${SOCMAKE_VERSION}/bootstrap/socmake-config.cmake" \
+  -o "${INSTALL_DIR}/socmake-config.cmake"
+
+curl -fsSL "${REPO_URL}/${SOCMAKE_VERSION}/bootstrap/socmake-config-version.cmake" \
+  -o "${INSTALL_DIR}/socmake-config-version.cmake"
+
+echo "✓ SoCmake bootstrap installed successfully! in ${INSTALL_DIR}"
+echo ""
+echo "To use in your CMakeLists.txt add the following:"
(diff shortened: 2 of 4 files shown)


mtravaillard:
Hi @Risto97,

I did the `curl` command, the 2 files were correctly added into `~/.local/lib/cmake/socmake`.

I took the *simple_sc_sv* example from SoCMake that I copied in another repo and changed how I get SoCMake.
It starts like this :
```
cmake_minimum_required(VERSION 3.25)
project(simple_sc_sv_example C CXX)

# Get SoCMake
find_package(SoCMake)

# Compile SystemC library during initial CMake configure time
systemc_build(VERSION 3.0.0 EXACT_VERSION)

# Add the Verilog DUT
add_subdirectory(adder)
    .
    .
    .
```

But unfortunately, I had to add `~/.local/lib/cmake/socmake` to my `PATH`, so when I did the configuration, it could find `socmake-config.cmake` and properly work. Otherwise, it stopped at systemc_build has it doesn't know this CMake command

Have I done something wrong? 
PS: In my PATH variable, there is only path to `/usr/...` and one to `~/bin`, nothing to `~/.local/...`

Risto97:
Hi @mtravaillard ,

Thanks for testing.
I see what is happening.

There is one feature of CMake that can help you debug the `find_...()` functions.
You can pass `-DCMAKE_FIND_DEBUG_MODE=ON` while configuring, or I prefer to wrap the specific find command like this, so I get less noise:
```cmake
set(CMAKE_FIND_DEBUG_MODE ON)
find_package(SoCMake 0.3.1 EXACT REQUIRED)
set(CMAKE_FIND_DEBUG_MODE OFF)
```
This will give you a lot of information about how CMake is looking for your package, and what variables influence the search.

What is happening for me is that I have the `~/.local/bin` on the `PATH` on every machine I use basically, so I didn't see that this affects the search algorithm.
When I removed it, it also doesn't find it for me.

I was hoping that there would be a user directory, that the CMake would always search in by default, but it seems like its not the case. This is what I get when for the search directories when I remove the `~/.local/bin` from the path:


```
find_package considered the following locations for SoCMake's Config
  module:

    /tools/work/tests/cmake/socmake_find/buil/CMakeFiles/pkgRedirects/SoCMakeConfig.cmake
    /tools/work/tests/cmake/socmake_find/buil/CMakeFiles/pkgRedirects/socmake-config.cmake
    /home/risto/go/SoCMakeConfig.cmake
    /home/risto/go/socmake-config.cmake
    /home/risto/.cargo/SoCMakeConfig.cmake
    /home/risto/.cargo/socmake-config.cmake
    /home/risto/SoCMakeConfig.cmake
    /home/risto/socmake-config.cmake
    /usr/lib64/ccache/SoCMakeConfig.cmake
    /usr/lib64/ccache/socmake-config.cmake
    /usr/local/SoCMakeConfig.cmake
    /usr/local/socmake-config.cmake
    /usr/SoCMakeConfig.cmake
    /usr/socmake-config.cmake
    /home/risto/.fzf/SoCMakeConfig.cmake
    /home/risto/.fzf/socmake-config.cmake
    /SoCMakeConfig.cmake
    /socmake-config.cmake
    /opt/SoCMakeConfig.cmake
    /opt/socmake-config.cmake

  The file was not found.
```

What would be a safest solution is to put it in `/usr/local/lib/cmake`, but I want to avoid need for root permissions for this.
So it seems like the only option is to ask users to add this into their `~/.bashrc` after the `bootsrap.sh` script finishes.
```
export CMAKE_PREFIX_PATH=~/.local/lib/cmake/:$CMAKE_PREFIX_PATH 
```

Well, fine I guess its not a big issue, that's how many  tools do it, (node, cargo, fzf, go, ...)

README.md
```diff
@@ -38,21 +38,23 @@ Getting started
 SoCMake is lightweight and has minimal dependencies.
 The only mandatory dependencies are `CMake>=3.25.0` and `make` ([Install Dependencies](https://hep-soc.github.io/SoCMake/docs/getting_started))
 
-There is no need to install SoCMake on your system, it is possible to fetch it in your `CMakeLists.txt`
+### One-Time Bootstrap Installation (Recommended)
 
-Create a file called `CMakeLists.txt`
+Install the SoCMake bootstrap files once on your system:
+**TODO Change the repo and branch name after merging**
```
benoitdenkinger, line 44: This PR looks good to me, don't you want to change this now?

bootstrap/bootstrap.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/sh
+set -e
+
+# Configuration
+SOCMAKE_VERSION="${SOCMAKE_VERSION:-bootstrap}"
+INSTALL_DIR="${HOME}/.local/lib/cmake/socmake"
+REPO_URL="${SOCMAKE_REPO_URL:-https://raw.githubusercontent.com/risto97/socmake}"
```
benoitdenkinger: Same here.

Risto97:
I changed it to HEP-SOC repository and also to the develop branch.
The problem is I want to fix it to a TAG, so that we make sure the bootstrap script does not need to be update in the future.

So lets maybe plan to make a TAG once we merge this and then update this script again to use that fixed tag instead of develop.

When do you expect we can drop a new tag?

benoitdenkinger:
Regarding the tag, this means we merge on master. I would like a few more things to be done, especially https://github.com/HEP-SoC/SoCMake/issues/119 and cleaning what is not used or not well documented and document the rest. Considering I don't have so much time before March, but now @mtravaillard is here to help too. So a good target would be end of March.

Risto97:
Ok thats fine, the bootstrap script should work pointing to develop, as long as we don't remove it from there.
So no urgency.

PR #179, opened 2026-01-29
Fix #175

### LINKED ISSUE #175: Problem with systemc build encountered with the simple_sc_sv example
When testing the simple_sc_sv example, I encountered a problem with SystemC not being correctly built (maybe ?) and creating problem when "verilating" the design.

I get a message error telling `SystemC requires a C++ compiler version of at least C++17`.

I tried with 2 versions, GNU 8.5 and GNU 12.1, which are both supporting C++17, the same issue happens in both case.

I'm using cmake 3.28.6 on red hat entreprise 8.10.
The weird thing, is that on my desktop, running also with cmake 3.28.6 but on alma linux 9.7, I don't have this problem...

I have been able to bypass this by putting `set (CMAKE_CXX_STANDARD 17)` in the `CMakeList.txt` of this example.

@Risto97, have you ever encountered this specific issue ? I don't know if it's something missing on the example or a problem coming from the setup...

Edit: On Red Hat Linux : verilator 5.035 (devel v5.034-7) from osscad-suite-20250226
On Alma Linux : verilator 5.044, i tried to run with verilator 5.034 but didn't got the problem either

mtravaillard:
Fixing issue #175 by setting the C++ standard to C++17, which is needed when using systemc and verilator together.
Depending on the compiler used, C++17 might not be the default standard.

### DIFF
--- a/examples/simple_sc_sv/CMakeLists.txt
+++ b/examples/simple_sc_sv/CMakeLists.txt
@@ -1,6 +1,9 @@
 cmake_minimum_required(VERSION 3.25)
 project(simple_sc_sv_example C CXX)
 
+# Verilator with SystemC requires C++17
+set(CMAKE_CXX_STANDARD 17)
+
 # Include SoCMake build system
 include("../../SoCMakeConfig.cmake")
 


Risto97:
You can merge this if you are ready

PR #181, opened 2026-01-30
Fix for #177: Changes on systemc/uvm-systemc build script to add a path for installation & build directory

### LINKED ISSUE #177: Problem when re-using a SystemC built by an other project,
When using a systemc built by another examples, we can find it but not use it correctly, there is an issue with the different useful path, especially to get the headers.

When calling SystemC build script, it will call `find_package()` to try to look for it, if not found, it will build it and everything will go fine.
In the case of finding it, depending on the environment, it might not work when trying to build anything with it. 

I used `CMAKE_FIND_DEBUG_MODE` like showed in this [PR](https://github.com/HEP-SoC/SoCMake/pull/174#issuecomment-3806838134).
I found out that when you build SystemC, it set a variable (global and cached), called `CMAKE_FIND_USE_PACKAGE_REGISTRY` which will give the path to the **building directory** which contains `SystemCLanguageConfig.cmake`. And another variable is set, `SYSTEMC_HOME` which give the path to the installation, but this one is not cached.

If for example, we want to use UVM-SystemC, we need SystemC before installing, so when looking for the package, we find it, and we set `SYSTEMC_HOME` at `{CMAKE_FIND_USE_PACKAGE_REGISTRY}/../` (as it's not cached, we don't have access to the value set at the building step).
In this case, the problem is that :
```
CMAKE_FIND_USE_PACKAGE_REGISTRY = .../build/systemc-build/src/systemc-build
SYSTEMC_HOME = .../build/systemc-build/src/
```
So it's looking for header in `.../build/systemc-build/src/systemc/include/` but the include folder does not exists...

So if I want to run UVM-SystemC, I can only do it by building systemc just before, as it will give the right value for `SYSTEMC_HOME`.

This is not really helpful for our CI, as we would have to build systemc all the time we have to use it.
It's even sometimes creating problem when trying to run the simple_sc_sv examples

@Risto97 have you ever encoutered this issue ?

mtravaillard:
In this PR, I modified the argument given in the build script to give the path to where we want to build the repository.
This path is now kind of "linked" with the installation path. This way, it's easier to find installation and building path of an already existing SystemC.
It solves the problem of trying to look for headers that are not existing, as explained in #177

The motivation of those changes, was to solve the issue with uvm-systemc and also find a way to be able to run examples using systemc on our private CI, without building it for every examples.

Now, by using `cmake ... -DFETCHCONTENT_BASE_DIR=/custom_path/` we are able to build and install SystemC where we want.

What do you think of this solution @Risto97 @benoitdenkinger ?


mtravaillard:
The commit 40547d3 fixed the last problems that I had.
it was only working in a few configuration/situation, now it works in any case, with or without SystemC built and with or without using `FETCHCONTENT_BASE_DIR`

Risto97:
Hello @mtravaillard ,

I tested this pull request and both examples build for me.
Both when it compiles from scratch or when it reuses an already existing library.
Tested on Fedora 43.
The only small issue is the warnings that get printed for `SYSTEMC_HOME`, ... variables, but I guess its work in progress.

I don't have much opinion on the build scripts, as I don't have them as part of any flow right now to test them.

mtravaillard:
Oh yes my bad, i forgot to remove the warning when committing the changes, thanks for telling me !

benoitdenkinger:
@mtravaillard 

Is this fixing problem on our gitlab runners as well? Can we have now clean separated build of systemc/uvm-systemc for the different examples?

mtravaillard:
Yes it does fix the problem that we had on our gitlab runners, it also fix a problem that would happen locally

PR #182, opened 2026-02-05
Task #171: Change cmake minimal version to 3.28

### LINKED ISSUE #171: Verilator and different minimal cmake version
When using Verilator, some variables are called and requires a newer version than cmake 3.25.

On the branch **master**: we are using [BUILD_JOB_SERVER_AWARE](https://github.com/HEP-SoC/SoCMake/blob/571d08365b55fea0f64fa2c2facecdfce58088ee/cmake/sim/verilator/verilator.cmake#L174), this option has been added in [cmake 3.28](https://cmake.org/cmake/help/latest/release/3.28.html)

On the branch **develop**: we are not using this option, but we are setting a policy, [CMP0144](https://github.com/HEP-SoC/SoCMake/blob/6c997dcf6706a7de1049e3cba8e7a64a9440d166/cmake/sim/verilator/verilator.cmake#L43), which has been added in [cmake 3.27](https://cmake.org/cmake/help/latest/policy/CMP0144.html)

So as develop should be merged in master in the future, should we also “upgrade” the minimal cmake version ?

mtravaillard:
Change cmake minimum required version to 3.28, to support features from versions 3.27 and 3.28 that were added in SoCMake

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 
-cmake_minimum_required(VERSION 3.25)
+cmake_minimum_required(VERSION 3.28)
 
 # Version is set first to reuse it across the SoCMake project
 # For example, docs/CMakeLists.txt includes this file to retrieve the version
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Getting started
 ---------------
 
 SoCMake is lightweight and has minimal dependencies.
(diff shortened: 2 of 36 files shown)


benoitdenkinger:
Could you remind me @mtravaillard the need for setting a higher minimal version?

mtravaillard:
It's when using verilator, as explained in issue #171, we are using things that were added in cmake 3.27 and 3.28

mtravaillard:
As said in the issue #171, it not needed to upgraded the cmake minimal version, using the policy was not needed so we can stay in cmake 3.25.
I will remove the policy in a new PR.

PR #184, opened 2026-02-08
Help menu with JQ

Risto97:
Currently the help menu tables are created with CMake language.
Although it works, its quite complicated and difficult to change the formatting.

Currently I have issue with help menus getting too wide, if the content of any of the columns becomes wide, and they wrap over the terminal screen.
The solution would be to limit the number of characters per column and truncate text, but this is another feature to add to already complicated CMake code for help menu's.

Instead I explored the possibility to use the [JQ language](https://jqlang.org/) for generating these tables.
JQ is a JSON processor that can transform JSON data into any text in an elegant way.

The printing of the table becomes pretty easy, as shown [`cmake/utils/help/ip.jq`
](https://github.com/Risto97/SoCMake/blob/0dae2460b5dfe9b8c8e435551afa2a6a89d19527/cmake/utils/help/ip.jq#L10-L22)

The idea is that in CMake we just dump the JSON file with the list of targets, ips, options, and from there JQ is processing the .json and creating a table.
This way we keep the CMake code simple and it only does the JSON dumping and target creation.
The JSON files look something like this:

```json
{
  "targets" : 
  [
    {
      "description" : "Build first-stage bootloader (ROM code)",
      "groups" : [ "build" ],
      "name" : "build_bootloader"
    },
...
```

```json
{
  "options" : 
  [
    {
      "advanced" : "FALSE",
      "current" : "ON",
      "default" : "ON",
      "description" : "Enable hardware floating-point unit",
      "name" : "ENABLE_FPU",
      "type" : "Boolean",
      "values" : [ "ON", "OFF" ]
    },
...
```

I also introduce the `SOCMAKE_GROUPS` property, that allows for placing the Targets, options, ips into groups, these groups can be used to create custom help menus, but maybe also for other purposes.
Another use of `SOCMAKE_GROUPS` is creating a custom target, that would depend on all targets of a group for example.

Having this JSON file, allows also for some further ideas, I am planning on creating a TUI interface (like lazygit, ccmake, ...) for SoCMake, to allow users to change the configuration options, run targets, list and browse IPs, ...
I already experimented here: [cmake-tui](https://github.com/Risto97/cmake-tui)

A downside is that this introduces JQ as optional dependency (if you want help menus). Although JQ is very widely available on most linux distros, and its very likely already present on many machines, as its a common dependency.

@benoitdenkinger , @mtravaillard Any opinions on this change?


benoitdenkinger:
I think adding an extra dependency is worth it, I remember doing it also at some point in CMake and it is a pain for the reasons you describe, so I completely in favor of a better solution. I'll give it a try

benoitdenkinger:
It is working well for me, I just had to update `SoCMakeConfig.cmake` with the new help_utils.cmake and groups.cmake

mtravaillard:
after updating `SoCMakeConfig.cmake` it worked fine for me. 
I just had 2 small things that i noticed, when using enumeration, it still troncate the list, for example, with simple_verilog example :
```
Option                         Type       Value           Default         Values                                   Description                                       
------------------------------------------------------------------------------------------------------------------------------------------------------
SIMULATOR                      Enum       iverilog        iverilog        [iverilog,vivado_sim,questa,xcelium,v... Which simulator to use                            
------------------------------------------------------------------------------------------------------------------------------------------------------
```
So we miss some information, I was able to have everything printed before with shell in full screen.
If i wasn't in full screen, it would have wrapped over, so the problem that you solved.
Maybe we could find a way, depending of the size the enum list take to print it fully and give less space to the other parts ?

And also, the `PRINT_ON_CONF` argument for help() is not working anymore on my side.

Risto97:
Yes, the enumeration values get also truncated which is not ideal.
What if we get rid of the Default value column for the Options menu?
This can give us more space for the enumeration values.

There is also an option to wrap the text into the next line, but then the scripts get a bit more complicated

mtravaillard:
yes i think we can get rid of the default value column in the help message. For me, having a full enumeration values is more useful than the default values.

Risto97:
I don't know if this is an overkill, but now its scalable based on terminal width by calling `tput cols`.
It is not responsive, it just scales at the time you call the help target.
I removed also the default value to make it more compact.

mtravaillard:
Scaling based on the size of the shell is a good idea especially when things are truncated.
But i think it adds a problem of readability in some case like shown in this screenshot, you have the target on one side of the shell and the dsecription on the other side instead of being close to it :
image
I also saw that it will still truncate the value even if it's possible to print them like on this case:
image
there is still some big empty space between the option and the type and it's not going to the end of the shell.

This is the only two "problems" that I encountered while trying it. Maybe it's small problems that have complicated solutions...
If it's too long to fix, honestly I think we can use the change you have already done, it's working really well with it and it's still very clear

Risto97:
Thanks for the feedback.

I think the trick is in playing with proportions of the columns a bit.
I tried to make it more aligned now, so it looks nicer.
The width of the names of targets, ips, and options are now aligned so it should look more organized.

We can have some more advanced techniques like finding the longest string in the column as we did before, but I am trying to avoid over complicating it.
Lets maybe try to use it in this form, and see if tweaking column widths is enough.

The second issue you saw was just that the separator `---` was not going to the end of the line for the options, so this should be fixed now.

mtravaillard (commented):
This is all good for me, it looks way better now and easier to read, it's a good upgrade.

PR #185, opened 2026-02-10
fix #171 : Remove policy CMP0144

### LINKED ISSUE #171: Verilator and different minimal cmake version

mtravaillard:
Policy CMP0144 was still set and was not needed anymore, since `BUILD_JOB_SERVER_AWARE` has been removed on the develop branch.
This way we are still compatible with cmake 3.25 as a minimum version and solving a cmake version issue that was not necessary anymore.

### DIFF
--- a/cmake/sim/verilator/verilator.cmake
+++ b/cmake/sim/verilator/verilator.cmake
@@ -38,14 +38,10 @@ function(verilator IP_LIB)
     ## Find verilator installation ###
     ##################################
     if(NOT VERILATOR_HOME)
-        # Ensure CMP0144 is set before find_package to get rid of warning
-        cmake_policy(PUSH)
-        cmake_policy(SET CMP0144 NEW)
         find_package(verilator REQUIRED
             HINTS ${VERISC_HOME}/open/* $ENV{VERISC_HOME}/open/*
             )
         set(VERILATOR_HOME "${verilator_DIR}/../../")
-        cmake_policy(POP)
     endif()
 
     find_file(_VERILATED_H verilated.h REQUIRED
--- a/cmake/sim/verilator/verilator/CMakeLists.txt
+++ b/cmake/sim/verilator/verilator/CMakeLists.txt
@@ -1,5 +1,4 @@
 cmake_minimum_required(VERSION 3.25)
-cmake_policy(SET CMP0144 NEW)
 
 project(${TARGET})
 


Risto97:
2026-02-10T21:04:57,508282948+01:00

The problem is that without this policy I am seeing this warning.
If I bump the `cmake_minimum_required(VERSION 3.28)` in the example and also `cmake/sim/verilator/verilator/CMakeLists.txt` then the warning is gone.

To get rid of it I think its either the policy or bump the CMake minimum version

mtravaillard:
if we keep the policy, we have to bump the minimum cmake version anyway

Risto97:
Are you not seeing this warning?
I tried updating to latest verilator and the issue is still there.
My versions are:

```
cmake version 3.31.6
Verilator 5.044 2026-01-01 rev v5.044
```

mtravaillard:
The policy `CMP0144` has been added in cmake 3.27, so has I tried with cmake 3.26.5 and 3.25.0, I had no warning, with verilator 5.044 and with 5.012 (built by an example).
I upgraded cmake and got the warning this time.

So we should keep the policy and upgrade the version to at least cmake 3.27 ?

Risto97:
I see, that makes sense, thanks for clarifying.

I would be fine with bumping the minimum version to `3.27` and keeping the policy, sorry for going back and forth on this issue.
Lets see also what @benoitdenkinger  thinks.

benoitdenkinger:
Good for me

mtravaillard:
No problem for going back and forth, at least now, we all understood the problem and agreed how to solve it !

PR #186, opened 2026-02-12
Change dpi-c example to keep only one verilator option

mtravaillard:
In the dpi-c option, there is 2 way of doing the example with verilator.
In one case, with use the MAIN argument and in the other one, it's done an other way to link the verilated design.
From my understanding, they do the same thing.

@Risto97, is it due to older version of the code or verilator that weren't supporting dpi properly ?
I saw that this simulator option has been added in a commit with change to verilator/dpi-c library support.
So I'm wondering if the other option add more information on how to use verilator with DPI or is just the "old way" of doing it..


Risto97:
Hello @mtravaillard ,

I think you are right, there was a moment where linking the DPI-C library to verilated design didn't work properly in SoCMake, so I made a small workaround for verilator.
Now it works properly we can keep only one.

Maybe there could be value of also showing how DPI-C library can be used when using Verilator with user supplied C++ main, but that is trivial so maybe there is no reason to have an example for that case.

In general, this pull request is ok for me.

PR #189, opened 2026-02-13
Upgrade for examples documentation

### LINKED ISSUE #183: Table for the examples
I have thought that a table like [this](https://github.com/mtravaillard/SoCMake/tree/99b00d3865c3740b8f7afbe472aa1166670bf87a/examples) could be helpful to have an overall view of the examples that can be run depending on someone's configuration.
It gives a fast and clear representation of the examples that are available, the tools that they support and their current state (more useful for us).

What do you think about this @Risto97 ?
I don't know if we should put it only here, put it only in the documentation or both ?

I haven't found a way to makes it directly compatible with docusaurus (yet?) so if we choose both, it would have to be a copy, so both will have to be updated every time..

mtravaillard:
The goal of this PR, is to improve the documentation, especially about the examples.

The following changes have been made : 
 - New pages for : options, mixed language, dpi-c, systemc, uvm-systemc and cocotb.
 - The table discussed in #183 have been added in the example directory and in the documentation.
 - The README to generate the documentation locally has been updated.
 - Few changes to the already documented examples, for tips/info or useful things.
 - The link to clone the repo in `getting_started.mdx` have been changed from gitlab to the public github repo
 
 @Risto97 @benoitdenkinger What you think about this update ?
 You can generate the documentation locally to see the change, if you ever see typo or anythings that need to be change, don't hesitate to tell me. I tried to proofread but you can always miss something :)
 
 To discuss : 
  - I don't know if it's better to put the `simulation` example first or keep the current organization ? I tried to do with table first, then some basic useful features of socmake and after different simulations and tools, but this examples is very complete and detailed it might be useful to add something to the name or else to "highlight" it...
  - About the video in this same example, I haven't been able to find a way to fix it, should we delete it or there is a way to make it work that I have missed ?


Risto97:
Hello @mtravaillard ,

Great initiative, we needed better documentation for a while now, finally someone stepped up to work on it.
I will help you when I find time in rewriting the current documentation, mostly about the backend of how SoCMake works.

> I don't know if it's better to put the simulation example first or keep the current organization ? I tried to do with table first, then some basic useful features of socmake and after different simulations and tools, but this examples is very complete and detailed it might be useful to add something to the name or else to "highlight" it...

It seems good this way, it feels like its progressing in complexity as a tutorial.

> About the video in this same example, I haven't been able to find a way to fix it, should we delete it or there is a way to make it work that I have missed ?

This is something I messed up in the beginning. I uploaded a video file to the repository which permanently bloated the repo by few megabytes. I would gladly remove it from the history, but that would mean we would need to modify the git history. I would maybe even be ready to take that risk, to cut these few MB of binaries from the repo.

mtravaillard:
yes it looks great like this ! 
I will also try to update/complete the API documentation now, there is almost 70% of it that is not done but I will do it in a new PR because it might be longer to do than this one.

> This is something I messed up in the beginning. I uploaded a video file to the repository which permanently bloated the repo by few megabytes. I would gladly remove it from the history, but that would mean we would need to modify the git history. I would maybe even be ready to take that risk, to cut these few MB of binaries from the repo.

Can't we just delete all the files and code related to it in a new commit ? It should be easier and less risky no ?

benoitdenkinger:
> Can't we just delete all the files and code related to it in a new commit ? It should be easier and less risky no ?

This will not delete the file from the history, just from the commit. A full rewriting of the history is needed to make it completely disappear.

I would be in favor of trying to rewrite the history, keeping the repo as small as possible is always good. All the doc images should also move to git lfs.

For now I'm merging this one.

PR #191, opened 2026-02-14
Add fzf_ninja function in the misc directory

Risto97:
Adding the `fzf_ninja.sh` file that allows for fuzzy searching through a list of ninja targets and on pressing enter it would autocomplete it in the terminal.

I currently but it in the `misc` directory, similar to how ninja stores bash-complete scripts and similar https://github.com/ninja-build/ninja/tree/master/misc

2026-02-14T16:47:09,908580508+01:00

### DIFF
--- a/misc/fzf_ninja.sh
+++ b/misc/fzf_ninja.sh
@@ -1,0 +1,15 @@
+# Fzf function for running ninja targets
+# Copy the content of this file to ~/.fzf.bash or ~/.bashrc
+# Change the keybinding to the using the `bind` command as shown below, by default its using Alt+n
+
+_fzf_ninja_target() {
+    local target
+    target=$(ninja -t targets all 2>/dev/null | cut -d: -f1 | fzf --height 40% --reverse)
+    if [ -n "$target" ]; then
+        READLINE_LINE="ninja $target"
+        READLINE_POINT=${#READLINE_LINE}
+    fi
+}
+
+# Bind to Alt+n
+bind -x '"\en": _fzf_ninja_target'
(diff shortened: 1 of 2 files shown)


Risto97:
Added 3 more functions that are more SoCMake specific.
They rely on `json` files dumped by `help()` functions.
This allows for having a bit more information in fzf script, and also constructing `cmake` configuration commands through FZF menus as shown in below screenshots.

There are 3 key bindings suggested by default:
* `Alt+i` - for displaying ips
* `Alt+o` - for displaying options
* `Alt+p` - for displaying targets

To use these functions source the `fzf_socmake.sh` file in your `.bashrc` or `.fzf-bash` file.

The dependencies are `fzf` and `jq`

I did use an LLM partially to write these, so use at your own risk.

# Target fzf menu

Allows to fuzzy search through targets, and if pressed enter the function would create a shell prompt:

```
ninja build_fpga_bitstream
make build_fpga_bitstream
```
Depending on the build system used

2026-02-14T19:52:34,315719156+01:00

# IP FZF menu

This one works the same as the Targets, but it doesnt make too much sense as you cannot execute IP target normally.

2026-02-14T19:51:26,605498807+01:00

# Options FZF menu

This can be very helpful and it automatically creates the `cmake -DSIMULATOR=modelsim ../` commands by going through FZF menu.
In case we have enumeration the available options are shown as FZF recommendations.
The options are appending to the current command in the prompt, so press `Alt+o` multiple times until you have all the options overriden.

2026-02-14T19:51:37,345886334+01:00
2026-02-14T19:51:57,953672238+01:00
2026-02-14T19:52:09,763381752+01:00

Risto97:
We could install this automatically for the user through the already existing bootstrap script.
We should of course ask it through the prompt if they want it to be installed or not, what do you think, is this useful?

benoitdenkinger:
I quickly checked it and it works for me on the help example. It is a nice feature that we could add to the bootstrap (asking the user indeed), but it also adds a dependency. For me this can be merged.

Issue #196, opened 2026-03-16
copy_rtl_files() fails with non absolute paths


mksoc:
Not sure how this was not observed before, but the current implementation of `copy_rtl_files.py` outputs only module names, which of course `vhier` cannot trace back to full path filenames.

Also, include files are returned as the filename only without the path. We need the full path to be able to copy them to the output folder.

benoitdenkinger:
`vhier` version used until now was latest commit on master: https://github.com/veripool/verilog-perl/tree/1188082772ebc4715797178128c1154e7be66e81

The function works with this version, but it seems this version (3.483) was never tagged properly, although it seems to be a valid release.

The changes proposed in #197 are actually a backward fix for v3.482

mksoc:
`vhier` will be removed in favor of `slang`

PR #197, opened 2026-03-16
Fix copy_rtl_files

mksoc:
https://github.com/HEP-SoC/SoCMake/issues/196


benoitdenkinger:
Are these changes working with v3.482 and v3.483?

PR #204, opened 2026-04-07
IP-XACT up-converting and more

mtravaillard:
Add Accellera's official script to up-convert files following the 2009 and 2014 ipxact standard.

An optional argument has been added to give the ipxact standard of a given file, to convert it to 2022 if needed and supported.

This is done at the configuration stage and the generated files will go in the build folder, so it stays clean for the user.

@Risto97 what do you think of this ?


Risto97:
Hello @mtravaillard ,

I had initially the same idea, I implemented more or less the same as you did.
However, I decided to make things simpler at the configure phase.

What do you think about creating a simple bash utility (or any scripting language) in a separate repository?

My solution was that if I have older IPXact files, I would just pre-convert them and track the 2022 files in git.
These conversion scripts are actually not perfect, and sometimes I still required manual edits from very old IPXact files (2009), making the automatic conversion less useful.

Let me know why you think it would be beneficial to do this as part of SoCMake, maybe I am missing something.

mtravaillard:
Hi @Risto97, i see what you mean.

>What do you think about creating a simple bash utility (or any scripting language) in a separate repository?

I was thinking that, it would be easier for an external/new user as it's more simple to use a function in cmake than running a script and then calling the function in cmake.

Maybe, we could store the 2022 files in the input directory and not in the build dir. This way, if it doesn't work, the user can modify it and give it has an input and can also track it with git if needed ? But if manual edits 

>These conversion scripts are actually not perfect, and sometimes I still required manual edits from very old IPXact files (2009), making the automatic conversion less useful.

Maybe if it requires manual edits too often, my idea will be a mess, but if it's rare i think it will be good too, tell me what you think, i guess that your idea would be better if we have to do them often.

Risto97:
I am not sure either what is the best way, as in the end it also comes down to preference.
The only reason I am hesitant about this is that we are adding quite a bit of files to maintain in SoCMake, the xslt scripts from IPxact.
But maybe its not a huge issue.

PR #209, opened 2026-04-27
Questasim fix exit codes

Risto97:
Questasim does not propagate the exit codes to the CLI after it executes the testbench.
It means that the calls like `$fatal(), $error()` are ignored by the simulator and not forwarded to the shell.

This leads to difficulties in CI flows as we cannot rely on the exit codes, even if the `$fatal()` was called the exit code can still be `0`.
This pull request should fix that by using a default `run.do` script instead of inline `tcl`.

I also added the `DO_FILES` argument to `questasim` so that user can pass any number of tcl scripts to execute in place of the default ones.

@mtravaillard can you please check if this breaks any of your flows (in case you use questa), or internal CI pipelines.

### DIFF
--- a/cmake/sim/siemens/questasim.cmake
+++ b/cmake/sim/siemens/questasim.cmake
@@ -37,9 +37,11 @@ include_guard(GLOBAL)
 # :type RUN_ARGS: string
 # :keyword FILE_SETS: Specify list of File sets to retrieve the sources from
 # :type FILE_SETS: list[string]
+# :keyword DO_FILES: Specify list of do files to run with the run command
+# :type DO_FILES: list[string]
 #]]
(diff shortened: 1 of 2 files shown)


mtravaillard (commented):
I don't use Questa but it worked correctly on the CI, for 3 different versions of the tool.

Do you think it would be useful to add the DO_FILES argument to other tools (vcs, xcelium ...) or it's not relevant ? Currently they don't use/need `do run -all` like we do with Questa.

Risto97:
The other simulators work differently, and I am not sure if its possible to pass tcl scripts when running.
I think its more of a peculiarity of mentor tools.

I will test it a bit more and merge

PR #211, opened 2026-04-30
Add questasim vopt stage to have 3 step flow

Risto97:


mtravaillard:
Hi @Risto97 
I have launched the CI and the dpi-c examples does not work on 3 different versions.
I am investigating a bit, seems that the problem is coming from the `-sv_lib` flag not being supported by vopt.
It's passed by `dpi_libs_args`.

edit: if i remove `${dpi_libs_args}` from the elaborate cmd, it works.

Risto97:
Yes, it was not working for me also when I tested it on bigger design.
Maybe the `${dpi_libs_args}` need to be passed only to vsim but the `${hdl_libs_args}` maybe should go to vopt but not vsim.

I also had an issue about it complaining aboug `work` lib, maybe we dont see that in our test cases as we always have unnamed library (work).

Needs more work anyways, feel free to push if you have any fixes

